PR #7, opened 2023-02-17
feat: translating templaters/base

benfdking:


benfdking:
Implemented elsewhere

PR #146, opened 2023-11-20
chore: add impl `Ref::match_segments`

gvozdvmozgu:


gvozdvmozgu:
Thinking of splitting it into three PRs: adding parsers, extending the ANSI dialect, and finally this test

PR #151, opened 2023-11-30
chore: add ansi tests

gvozdvmozgu:


gvozdvmozgu:
To achieve this, we require the 'bracket_sensitive_look_ahead_match' function - implementation started in PR #152.

gvozdvmozgu:
@benfdking, let's merge this.
It seems there is a bug in the implementation of AnyNumberOf; I'm inclined to think it's better to fix it in a different PR. Haven't found the essence of the bug yet, but it's hindering further parser development.

PR #159, opened 2024-02-07
feat: move lib into its own folder

benfdking:


benfdking:
@gvozdvmozgu You ok with this change? Trying to create a library and a CLI and then build it in the CI.

gvozdvmozgu (approved):
LGTM

PR #162, opened 2024-02-09
feat: adding basic scaffolding for clap

benfdking:


crates/lib/Cargo.toml
```diff
@@ -1,15 +1,15 @@
 [package]
-name = "sqruff"
+name = "sqruff-lib"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
```
nikolaevx, line 6:
```suggestion
[lib]
doctest = false
```

gvozdvmozgu (changes requested):

crates/lib/Cargo.toml
nikolaevx: It seems that failures occur because Rust interprets any code in a code block as a doc test.

PR #180, opened 2024-02-25
chore: tests for the depth_map module

gvozdvmozgu:


benfdking:
@gvozdvmozgu what should we do with this one?

PR #216, opened 2024-04-12
chore: add descriptions to all rules

### LINKED ISSUE #215: Add descriptions to the rules in a callback method so we can generate the description page of them eventually
It would be useful for the rules to also have the function 

```
trait Rule {
  fn get_description() -> &str 
}
```

That contains the comment the Python code contains so we can use it to generate docs in the future.

gvozdvmozgu:
Close #215


gvozdvmozgu:
@benfdking

benfdking (approved):

PR #224, opened 2024-04-22
feat: parrallel implementation

benfdking:


louisjoecodes:
Timed:
- Folder with 315 duplicates of a random .sql file called  `test.sql` -> `SELECT * FROM test`
 - `sqruff lint  6.57s user 0.86s system 97% cpu 7.648 total`
 - `sqlfluff lint --dialect ansi  3.79s user 0.31s system 95% cpu 4.285 total`

Is it possible that the changes don't actually parallelise the listing process? The run method for each iterator is updated but thei iterator still processes each file sequentially I think.

Perhaps we need to parallelise the ParallelRunner?

PR #235, opened 2024-04-25
chore: less cloning and caching

gvozdvmozgu:


gvozdvmozgu:
@louisjoecodes, this should improve performance, can you check if it has an effect on your data?

gvozdvmozgu:
@louisjoecodes, this should improve performance, can you check if it has an effect on your data?

PR #257, opened 2024-05-01
chore: add rule `RF03`

gvozdvmozgu:


louisjoecodes:
Great work with this rule

PR #262, opened 2024-05-04
chore: Add CV03 rule for trailing commas in select clause

gvozdvmozgu:


crates/lib/src/rules/convention/CV03.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::utils::functional::context::FunctionalContext;
+
+#[derive(Debug, Clone)]
+pub struct RuleCV03 {
+    select_clause_trailing_comma: String,
+}
+
+impl Default for RuleCV03 {
+    fn default() -> Self {
+        RuleCV03 { select_clause_trailing_comma: "reuired".to_string() }
```
benfdking: I am going to guess this is a typo 

```suggestion
        RuleCV03 { select_clause_trailing_comma: "required".to_string() }
```

crates/lib/src/rules/convention/CV03.rs
```diff
@@ -0,0 +1,182 @@
...
+    use crate::core::rules::base::{Erased, ErasedRule};
+    use crate::rules::convention::CV03::RuleCV03;
+
+    fn rules() -> Vec<ErasedRule> {
+        rules_with_config("require".to_owned())
+    }
+
+    fn rules_with_config(select_clause: String) -> Vec<ErasedRule> {
+        vec![RuleCV03 { select_clause_trailing_comma: select_clause }.erased()]
+    }
```
benfdking, line 119: nit: Given we missed the type, I wonder if it is worth having a test that catches the "default case".

benfdking (changes requested):
Exciting work!

benfdking (approved):

PR #263, opened 2024-05-04
chore: Add Jinja templating support

gvozdvmozgu:


louisjoecodes:
Hey @gvozdvmozgu I think we can park JJ01/Jinja support for now. 

My 2 main concerns that we are only realising now:
- Means we need to add the Jinja dependency into our code-base which we want to avoid for now
- We don't fully know the compatibility between 'minijinja' and actual jinja which might lead to a risk further on.

Feel free to skip the JJ01 rule  for now, we will park this PR for now so feel free to skip. Great work nonetheless!

PR #274, opened 2024-05-13
ci: add benchmark to ci

benfdking:


benfdking:
@gvozdvmozgu I feel like this happened after BigQuery change? Any ideas?

crates/cli/src/bin/bench.rs
```diff
@@ -6,11 +6,13 @@ fn main() {
 
     let start = Instant::now();
 
-    let output = Command::new("target/release/sqruff.exe")
+    let output = Command::new("target/release/sqruff")
         .args(["lint", ".", "crates/lib/test/fixtures/dialects/ansi"])
```
nikolaevx:
```suggestion
        .args(["lint", "crates/lib/test/fixtures/dialects/ansi"])
```

crates/cli/src/bin/bench.rs
```diff
@@ -6,11 +6,13 @@ fn main() {
...
         .output()
         .expect("Failed to execute process");
 
+    assert!(output.status.success());
+
```
nikolaevx:
```suggestion
```

gvozdvmozgu (changes requested):

PR #285, opened 2024-05-14
feat: add publish to cargo step on release

louisjoecodes:
- added workflow_dispatch so that we can manually trigger the publish step
- Secrets stored.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -154,7 +154,6 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
       - name: Publish sqruff-lib to Cargo
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -2,6 +2,7 @@ on:
   push:
     tags:
       - "v*.*.*"
+  workflow_dispatch:
```
louisjoecodes, line 5: Adding this so that we can manually trigger the release workflow

.github/workflows/release.yml
```diff
@@ -145,3 +145,24 @@ jobs:
...
+  publish-to-cargo:
+    name: Publish to Cargo
+    runs-on: ubuntu-latest
+    needs: [release, check-versions-match]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
```
benfdking: I would remove this override as we don't use stable in sqruff. 

```suggestion
```

benfdking (changes requested):
Small change on toolchain

benfdking (approved):
Send it!

PR #294, opened 2024-05-14
chore: `String` -> `SmolStr`

gvozdvmozgu:


benfdking:
What's the reasoning behind this change? I was assuming a small performance bump but couldn't measurably see one. I guess the immutable makes it nice but I guess balancing that with extra dependencies. Curious about your thoughts?

benfdking (approved):

PR #298, opened 2024-05-14
chore: update capitalisation rules, handle missing rule in reference map, and support "sqruff" section in config

gvozdvmozgu:


benfdking (approved):

benfdking:
nit: can we mark these as `feat:` rather than `chore:` as they are more of a new feature than a chore e.g. updating a dependency etc.

PR #310, opened 2024-05-15
feat: add .sqruff to filename_options in ConfigLoader

gvozdvmozgu:


benfdking (approved):

benfdking:
@gvozdvmozgu can we repush this commit with prefix `feat:` so we can categorise it and talk about it correctly?

PR #314, opened 2024-05-16
chore: removing unusued code

benfdking:


benfdking:
@gvozdvmozgu I presume this test can go in the bin?

gvozdvmozgu:
Either delete it or replace it in the test with rule LT01.

PR #323, opened 2024-05-22
refactor: replaced `AHashMap` and `AHashSet` with `IntMap` and `IntSet` for `stack_hash_set` and `stack_positions` in DepthInfo

gvozdvmozgu:
…t` for `stack_hash_set` and `stack_positions` in DepthInfo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -678,6 +678,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "nohash-hasher"
+version = "0.2.0"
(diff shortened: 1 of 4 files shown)


benfdking (approved):

benfdking (approved):
If passes should this also be faster?

gvozdvmozgu:
This replaces the hash function with a noop. We are simply storing segment hashes, so we don't need to hash the hash again.

PR #324, opened 2024-05-22
ci: check for diffs after running tests

benfdking:


benfdking:
@gvozdvmozgu helps us avoid little mistakes like  https://github.com/quarylabs/sqruff/pull/325

PR #326, opened 2024-05-23
refactor: Use Arc for inner field in TemplatedFile struct

gvozdvmozgu:


benfdking:
@gvozdvmozgu nice benchmark!

louisjoecodes:
Hey @gvozdvmozgu is this change purely for performance? Given the benchmark above (nice!) I think we might not want to merge this in (unless this is change is part of something bigger).

gvozdvmozgu:
PositionMarker contains TemplatedFile, which is quite expensive to clone since it stores things like SQL source code and much more. During parsing, we don’t clone PositionMarker that often, but in the fix command, there are many instances because we need to calculate PositionMarker for new segments. Currently, the Segment interface only allows cloning of PositionMarker. This also affects linting, but not to the same extent. In this regard, we need to extend the benchmarks to cover the lint and fix commands, not just parsing.

PR #367, opened 2024-06-10
refactor: improve ParseContext structure and caching mechanism

gvozdvmozgu:


crates/lib/src/helpers.rs
```diff
@@ -177,3 +178,9 @@ fn with_ctx(f: impl FnOnce(&mut Vec<String>)) {
     }
     CTX.with(|ctx| f(&mut ctx.borrow_mut()));
 }
+
+pub fn next_cache_key() -> u32 {
+    static ID: AtomicU32 = AtomicU32::new(1);
+
+    ID.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |id| id.checked_add(1)).unwrap()
+}
```
benfdking: It might be worth documenting why this starts at one and its relationship to non-code segments.

benfdking (changes requested):
Nice work! I do think it's worth documenting the magic numbers as I had to go digging to understand your choices.

benfdking (approved):

PR #370, opened 2024-06-15
feat: implement new node matcher API

gvozdvmozgu:


gvozdvmozgu:
For this, we need a separate counter. We also have a problem where the `cache_key` is created for each instance, but currently, we don't store the instances; they are created and immediately dropped. This is about grammar, and it is resolved in [this pull request](https://github.com/quarylabs/sqruff/pull/370).

gvozdvmozgu:
Oh, I wanted to comment on a different PR

crates/lib/src/core/parser/segments/base.rs
```diff
@@ -734,7 +724,12 @@ impl Eq for ErasedSegment {}
 
 impl Hash for dyn Segment {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        self.type_name().hash(state);
+        // TODO: remove hack
```
benfdking: I think this can be removed?

benfdking (changes requested):
Amazing work! I think there's one TODO that we can clean up? If not let me know.

benfdking (approved):

PR #411, opened 2024-06-24
feat: remove all uuid

benfdking:


crates/lib/src/core/parser/grammar/base.rs
```diff
@@ -272,7 +271,7 @@ impl Anything {
 }
 
 impl Segment for Anything {
-    fn get_uuid(&self) -> Uuid {
+    fn get_uuid(&self) -> u32 {
         todo!()
     }
```
nikolaevx: We can remove this method.

crates/lib/src/core/parser/segments/base.rs
```diff
@@ -509,8 +508,8 @@ pub trait Segment: Any + DynEq + DynClone + Debug + CloneSegment + Send + Sync {
     }
 
     /// Group and count fixes by anchor, return dictionary.
-    fn compute_anchor_edit_info(&self, fixes: &Vec<LintFix>) -> AHashMap<Uuid, AnchorEditInfo> {
-        let mut anchor_info = AHashMap::<Uuid, AnchorEditInfo>::new();
+    fn compute_anchor_edit_info(&self, fixes: &Vec<LintFix>) -> AHashMap<u32, AnchorEditInfo> {
+        let mut anchor_info = AHashMap::<u32, AnchorEditInfo>::new();
```
nikolaevx: We can try replacing it with `IntMap` now.

gvozdvmozgu (approved):

crates/lib/src/core/parser/segments/base.rs
benfdking: My worry was overflow problems? What do you think? Not to do with the map but with u32?

nikolaevx: yes, because we are using a global counter, there is a chance of overflow. In my view, it would be ideal to have a local context for the file where we could store the counter, arena, and so on. i’m not sure if there are such large SQL codebases, but this issue could be bypassed by using a thread-local variable so that each thread has its own counter. However, this won't solve the problem for the VSCode plugin

benfdking: Yep I think local file counter would be better. You could then also make it local to a thread and then doesn't need to be atomic?

nikolaevx: For thread-local storage, we could use `RefCell<u32>`. For the method with passing context, I think a simple `Cell<u32>` or just `u32` would be sufficient.

benfdking:
@gvozdvmozgu Thinking about this one. I am seeing the bug in the web extension where you hit the u32 limit. How do you think we best solve this?

gvozdvmozgu:
I can't think of a simple solution right now

gvozdvmozgu:
After working on https://github.com/quarylabs/sqruff/pull/364, it occurred to me that we could simplify the process following this PR. Since the creation of segments is isolated in a separate phase, we can implement a local counter.

PR #474, opened 2024-07-09
feat: implementing the CV01 rule

benfdking:


benfdking:
@gvozdvmozgu when you have time, do you know why the memory isn't working here?

gvozdvmozgu:
It seems that the version of the grammar we are using incorrectly parses the expression.

The Python version (on which our version is based) and our version both produce an error:
```
== [pg.sql] FAIL
L:   1 | P:   1 | LT09 | Select targets should be on a new line unless there is
                       | only one select target.
                       | [layout.select_targets]
L:   1 | P:  33 |  PRS | Line 1, Position 33: Found unparsable section: '== 1 AND
                       | b != 2 AND c <> 3'
WARNING: Parsing errors found and dialect is set to 'ansi'. Have you configured your dialect correctly?
```

benfdking:
@gvozdvmozgu cheers! good catch, should be done now.

gvozdvmozgu:
It looks like the `default_config.cfg` needs to be updated.

benfdking:
@gvozdvmozgu nice find!

gvozdvmozgu (approved):

PR #489, opened 2024-07-11
feat: add  `ST07` rule

gvozdvmozgu:


crates/lib/src/rules/structure/ST07.rs
```diff
@@ -0,0 +1,244 @@
...
+    ON table_a.id = table_b.id
+```"
+    }
+
+    fn groups(&self) -> &'static [RuleGroups] {
+        &[RuleGroups::All, RuleGroups::Structure]
+    }
+
+    fn eval(&self, context: RuleContext) -> Vec<LintResult> {
+        if context.dialect.name == "clickhouse" {
```
benfdking, line 74: nit: This feels a bit off here. It's worth maybe adding a FIXME/TODO, as it is something we may want to do more uniformly in the future. 

```suggestion
        // TODO Figure out how to make this cleaner and part of the definition rules
        if context.dialect.name == "clickhouse" {
```

benfdking (approved):

PR #505, opened 2024-07-13
experiment to see rc arc performance

benfdking:
- **feat: enabling cv05**
- **chore: test of performance moving to rc versus arc**


gvozdvmozgu:
As far as I remember, we use Arc here so that Dialect can implement the Send trait.

PR #517, opened 2024-07-16
feat: implementing cv06

benfdking:
- **temp**
- **temp**
- **tempt**
- **temp**


gvozdvmozgu:
@benfdking

benfdking:
Any chance you can try to explain what you did? My understanding is that it was "seeking" through the file multiple times and thus you ended up with "multiple" identical fixes?

gvozdvmozgu:
There were errors in the arguments that were passed, and some conditions were inaccurate.

As far as I remember:
- If self.require_final_semicolon was active, segment was passed instead of context.segment.
- If it was not empty, a lint was created even though it consisted of comments or spaces.
- There was an incorrect selection of segments (it seems that in before_code, there was just is_code instead of !is_code).

I don't remember much else.

PR #564, opened 2024-07-29
feat: add new line at the end message

benfdking:


gvozdvmozgu (changes requested):
It looks like the UI tests need to be updated

gvozdvmozgu (approved):

PR #567, opened 2024-07-30
fix: implement of test that should work

benfdking:


gvozdvmozgu:
Do we have more test cases like this? I'm not confident in the current solution.

benfdking:
@gvozdvmozgu 

Nope, these are all test cases; I wrote them myself. I am happy to write more. Any particular reason for the nervousness, any edge case you think would not work?

gvozdvmozgu:
I think we can merge it now.

PR #594, opened 2024-08-07
Think about whether if you specify file directly it should not ignore it

benfdking:


crates/cli/tests/ui/hql_file.stderr
```diff
@@ -0,0 +1,2 @@
+The linter processed 0 file(s).
+All Finished
```
benfdking: Just ignored fully specified path

benfdking (commented):
Something probably we should fix.

PR #597, opened 2024-08-08
refactor: remove uuid

gvozdvmozgu:


crates/lib/test/fixtures/dialects/bigquery/dateparts.yml
```diff
@@ -245,7 +245,7 @@ file:
             - function_name_identifier: EXTRACT
           - bracketed:
             - start_bracket: (
-            - date_part: DATE
+            - warehouse_size: DATE
```
benfdking: `warehouse_size` here feels odd, but maybe I have misunderstood the Select context.

benfdking (approved):
What's not to love? 🚀  

- Delete 4k lines
- Made faster
- In theory 1 fewer dependency
- Most of it seems to have made the parsing better

There's one little thing going through the results where I questioned the `warehouse_size` nomenclature, but let me know if I am wrong. I have approved it so if you think I am wrong just move on 👍

PR #599, opened 2024-08-09
refactor: move `pprof` to dev dependencies

gvozdvmozgu:


benfdking:
I was wondering. Do we use serde in the actual CLI or is it just for the testing?

benfdking (approved):

gvozdvmozgu:
Serde is currently only used for testing. In the code, we'll need to hide the serialization code for the segment tree in TupleSerialisedSegment behind a cfg flag if we want to keep serde in the dev dependencies

PR #604, opened 2024-08-12
refactor: make `SyntaxSet` non-Copy and minimize cloning

gvozdvmozgu:


crates/lib/src/core/parser/segments/base.rs
```diff
@@ -167,33 +169,31 @@ impl ErasedSegment {
...
+        if !self.descendant_type_set().intersects(types) {
             return acc;
         }
 
         if recurse_into || !matches {
             for seg in self.segments() {
-                if no_recursive_types
-                    .map_or(true, |no_recursive_types| !no_recursive_types.contains(seg.get_type()))
-                {
+                if no_recursive_types.is_empty() || no_recursive_types.contains(seg.get_type()) {
```
nikolaevx:
```suggestion
                if no_recursive_types.is_empty() || !no_recursive_types.contains(seg.get_type()) {
```

PR #637, opened 2024-08-19
feat: add optional config file path to CLI

### LINKED ISSUE #636: Add ability to specify config file path
Hi, 
Is it possible to get the equivalent of 

```
sqruff fix --config="./path/to/my/.sqlfluff"
```

This functionality currently exists in sqlfluff and can be very handy if your config dot files are in a separate directory. 
From what I have tried the default functionality seems to be that it picks a config file it it is available in the root directory of where the cli is called from

gvozdvmozgu:
fix #636

### DIFF
--- a/crates/lib/src/core/config.rs
+++ b/crates/lib/src/core/config.rs
@@ -314,7 +314,6 @@ impl ConfigLoader {
         head.chain(tail)
     }
 
-    #[allow(unused_variables)]
     pub fn load_config_up_to_path(
         &self,
         path: impl AsRef<Path>,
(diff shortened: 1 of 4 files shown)


crates/lib/src/core/config.rs
```diff
@@ -318,21 +318,16 @@ impl ConfigLoader {
     pub fn load_config_up_to_path(
```
benfdking, line 317: nit: I think we can remove the unused variables in the above?

benfdking (approved):

PR #640, opened 2024-08-19
chore: add `web-dev` profile

gvozdvmozgu:


gvozdvmozgu:
okay, it seems like wasm-pack doesn't support custom profiles

PR #655, opened 2024-08-30
feat: implementing the placeholder templater

benfdking:


gvozdvmozgu (approved):
It looks good, but for the next PRs, we should:

- Update the linter/lexer, as it seems to panic when using templates.
- Transition from dynamic dispatch to `enum_dispatch` (similar to what we did with `Crawler`).

panic message
```
thread 'main' panicked at crates\lib\src\core\parser\lexer.rs:564:17:
not implemented
note: run with RUST_BACKTRACE=1 environment variable to display a backtrace
```

benfdking:
@gvozdvmozgu What did you run to get the above?

gvozdvmozgu:
config:
```
[sqlfluff]
dialect = ansi
templater = placeholder

param_regex = __(?P<param_name>[\w_]+)__
my_name = john
```

benfdking:
Ok I will implement this bit after.

PR #672, opened 2024-09-03
chore: update release.yml

gvozdvmozgu:


benfdking:
I think we should keep this step. The reason why is that we hit some errors in the past where if someone else was running the Github Action `install_sqruff` job during a release, you could get a failed job because it was looking for the latest release and it wouldn't have all the required binaries.

PR #687, opened 2024-09-04
internal: use mimalloc on windows

gvozdvmozgu:


benfdking (approved):
Nice! Would it be worth compiling and testing our code on Windows as well at PR time:?

PR #696, opened 2024-09-07
feat: add athena

benfdking:
- **feat: implement the trino dialect**
- **feat: implement athena dialect**


gvozdvmozgu:
@benfdking

PR #729, opened 2024-09-12
feat: web playground with WASM

gvozdvmozgu:


gvozdvmozgu:
Oh, looks like I accidentally installed classnames locally in the project root, and everything worked just fine! :)

Issue #740, opened 2024-09-13
Readme does not specify how to configure Sqruff


concur1:
The readme and documentation folder do not mention how to configure the dialect. From experimenting it seems the configuration is inherited from a .sqlfluff file. However this is not explicit or obvious to someone reading the readme for the first time. I suggest:
- A section is added to the readme that explains configuration can be done with a .sqlfluff file which should be mostly compatible with peoples existing .sqlfluff setup.

PR #745, opened 2024-09-13
Include information of default dialect and dialect configuration

concur1:
Added information on the dialect and dialect configuration to the readme.

related issue: #740

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1038,6 +1038,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "line-index"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "67d61795376ae2683928c218fda7d7d7db136fd38c06b7552904667f0d55580a"
+dependencies = [
+ "nohash-hasher",
+ "text-size",
+]
+
+[[package]]
 name = "lineage"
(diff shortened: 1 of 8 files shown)


gvozdvmozgu:
You also need to update the `README.MD` file located at `editors/code`.

benfdking:
```shell
cp README.md editors/code/README.md
```

For the update

gvozdvmozgu:
I think we should inform users that they need to use 'sqruff' as the config key names.

concur1:
> I think we should inform users that they need to use 'sqruff' as the config name and for the config key names.

So change:
```toml
[sqlfluff]
dialect = SQLite
```

to 
```toml
[sqruff]
dialect = SQLite
```
?

gvozdvmozgu:
Yes, and also the file name. We have case-sensitive matching for the dialect name, so SQLite won't work.

concur1:
Is Sqruff compatible with all Sqlfluff config files? Should I add a note around being able to copy over a config file from Sqlfluff?

gvozdvmozgu:
We strive for backward compatibility with SQLFluff; some things might not work, but overall, yes.

benfdking:
Thanks for the contribution

concur1:
np, happy to help =)

PR #746, opened 2024-09-13
Update README.md

gvozdvmozgu:


benfdking (approved):

gvozdvmozgu:
@benfdking  We might be able to use symbolic links to point editors/code/README.md to the root README.md.

benfdking:
I think I tried before and there was an issue with the vs code extension packaging.

Issue #739, opened 2024-09-13
Python support/installation with python


concur1:
First of all Sqruff is a very cool project so thank you!

I would like Sqruff to be made available as a python package so that I am able to install Sqruff using a python package manager such as PIP, Poetry  or UV. 

Python support would make using Sqruff easier for many python based projects as:
1. Some Python based teams will only have access to python packages and no ability to install packages to the OS.
2. Having the Sqruff version be defined in a pyproject.toml file makes it much easier to set up the environment of a project. Currently manual steps would be required.

Similar Rust based project such as Ruff and Polars are made available as a python package and I think this really helps their reach.

benfdking:
Hey @concur1, thanks for all the fantastic feedback! We'll get to it 🚀

concur1:
No problem!

I am happy to try and pitch in and help a bit with some of the things I suggested. So if you are happy with some of these suggestions and mark them ass accepted  I can will make a PR.

benfdking:
Feel free to open PRs for the suggestions and we'll review.

benfdking:
This is being deployed now with version 0.18.0 so we will have to see if it works.

benfdking:
@concur1 pipeline has run so if you want to try it, let us know.

concur1:
> @concur1 pipeline has run so if you want to try it, let us know.

Thanks, Yep, I have tested it on Ubuntu and it works fine.

gvozdvmozgu:
Need to remember to document that this installation method exists.

PR #747, opened 2024-09-13
Add python package CI [testing]

concur1:
Testing out a CI where the sqruff CLI will be packaged as a python .whl

This will probably need a bit of testing, it probably won't be ready for review for a while

related issue: #739


gvozdvmozgu:
Are you having any problems?

PR #760, opened 2024-09-16
Add python packaging (maturin)

concur1:
This PR includes the pyproject.toml file and github actions workflows needed to build and release the Sqruff CLI to PYPI as a python package.

Currently the workflows run on PR push. Eventually I think we want these workflows to run on release but it is easier to play around/test the workflows when they are triggered on PR push.

As I have created the [sqruff package in PYPI ](https://pypi.org/project/sqruff/) you should now be able to install the python package with:
```python
python -m pip install sqruff
```

I am happy to either delete this PYPI entry or hand the ownership over to you/quarylabs. 

Here are some things that probably need to be done (some of it on your end):
- Create a PYPI account that you own.
- Take handover from me of the sqruff PYPI entry.
- Create a PYPI token and put it in a github secret in the production environment - this is how I have it set up in my repo - using the production environment allows us to control who has the authority to trigger publishing to PYPI (this is needed to fix the failing publish job)
- Change the workflow from being triggered on push to being triggered on release.
- Add instructions to the readme around python installation.


benfdking:
Ok, so I have pypi account. Username is benfking

concur1:
> Ok, so I have pypi account. Username is benfking. Also thank you so much for this awesome work!

Ok, I have invited you as an owner

benfdking:
Have accepted! I am just finishing a few PRs off and then I will build it!

PR #769, opened 2024-09-16
feat(playground): add lineage panel

gvozdvmozgu:


gvozdvmozgu:
Currently, it's just a text version, and as I understand it, React Flow can be used for rendering. We also need to figure out how to allow the user to specify the column.

benfdking:
I think let's merge this for the time being and I will have a look at the React Flow as we already do some of this in `quary`

benfdking (approved):

PR #781, opened 2024-09-18
internal: revert use of pnpm for VS Code extension

gvozdvmozgu:


benfdking:
I fixed this in a slightly different way https://github.com/quarylabs/sqruff/pull/783

PR #788, opened 2024-09-18
fix(lineage): properly handle `ON` clauses in `JOIN` operations

gvozdvmozgu:


benfdking:
Is it worth adding a test case for this?

benfdking (approved):

PR #800, opened 2024-09-22
refactor: simplify runner

gvozdvmozgu:


gvozdvmozgu:
@benfdking  Could you check how it affects performance locally?

PR #801, opened 2024-09-22
refactor: use static dispatch instead of dynamic for `Matchable`

gvozdvmozgu:


gvozdvmozgu:
Once we eliminate let_chains, we'll be able to switch to stable Rust.

PR #871, opened 2024-10-11
feat: implement python templater

benfdking:
**To do**

* [x] Add proper python description
* [x] Make all the tests work in Python
* [x] Make it work in a GitHub action 
  * [x] cargo test with UI python
  * [x] make ci python
* [x] Make it a feature in lib in CLI that is by default off
* [x] Clean up crates/lib/src/core/linter/core.rs to not need the templaters directly..

**To do in next PR**

* [ ] Make sure I can be packageable (possibly on pip install)


crates/cli/Cargo.toml
```diff
@@ -25,12 +25,20 @@ bench = false
...
+features = ["python"]
+
 [features]
 codegen-docs = ["clap-markdown", "minijinja", "serde"]
+python = ["sqruff-lib/python"]
 
 [dependencies]
 sqruff-lib.workspace = true
 sqruff-lsp.workspace = true
+sqruff-lib-core.workspace = true
```
benfdking: Remove this line

crates/lib/src/templaters/python.rs
```diff
@@ -0,0 +1,249 @@
...
+#[derive(Default)]
+pub struct PythonTemplater;
+
+impl Templater for PythonTemplater {
+    fn name(&self) -> &'static str {
+        "python"
+    }
+
+    fn description(&self) -> &'static str {
+        "Python templater"
```
benfdking: Add proper description

crates/lib/src/templaters/python_templater_test.py
```diff
@@ -0,0 +1,577 @@
+"""Tests for templaters."""
+
+import logging
```
benfdking, line 3: Make tests work

PR #920, opened 2024-11-01
ci: fix maturin rust toolchain

concur1:
It looks like the python deployments have been failing because the rust-toolchain wasn't set and has been using `stable` by default when I think we want to use `nightly`.

This PR should fix that. I have tested in a previoud commit by changing the actions trigger to be 'pull request' and it built the artifacts fine for all the jobs: https://github.com/quarylabs/sqruff/actions/runs/11627091156

### DIFF
--- a/.github/workflows/python-ci.yaml
+++ b/.github/workflows/python-ci.yaml
@@ -6,7 +6,9 @@
 name: CI
 
 on:
-  pull_request:
+  push:
+    tags:
+      - "v*.*.*"
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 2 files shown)


gvozdvmozgu:
I don't think we need to use nightly. We should either specify version 1.81.0 in rust-toolchain.toml so it installs automatically, or set this version directly in the actions.

concur1:
Ahhh yes that makes more sense!

I have removed the references in the version in the workflow and updated the `rust-toolchain.toml` file. You can see that the workflow is now passing: https://github.com/quarylabs/sqruff/actions/runs/11627651026?pr=920

I will update the python-ci workflow to run on release instead of PR (which I was doing for testing).

gvozdvmozgu (approved):

concur1:
HI @gvozdvmozgu I think it is ready now. I don't think I can merge but once it's in I would really appreciate if you or @benfdking could run a release.

benfdking:
Let me have a look at this. This breaks my development flow for some reason can't run `cargo` when I change this.

benfdking (approved):

benfdking:
Thanks! Nice work.

PR #972, opened 2024-11-23
chore: add jinja cli templater test

benfdking:
Things to do: 

- [x] Make sure the python module gets the files inside of it without copying to venv
- [x] Clean up print statements
- [x] See how you would do it with installing the jinja and regex

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -85,10 +85,13 @@ jobs:
       - uses: actions/checkout@v4
       - id: setup-python
         uses: actions/setup-python@v5
+      - name: Create and activate virtual environment
+        run: |
+          python -m venv .venv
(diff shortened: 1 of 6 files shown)


benfdking:
Hey @mmitya yep! It's slightly fiddly but getting there

mmitya:
> Hey @mmitya yep! It's slightly fiddly but getting there

Hi. Do you have any plans to support DBT?

benfdking:
@mmitya, we might get there ;)

crates/lib/Cargo.toml
```diff
@@ -34,10 +38,12 @@ name = "depth_map"
 harness = false
 
 [features]
+default = ["python"]
```
benfdking: This needs to be reverted

PR #999, opened 2024-11-29
fix: github actions detection not working properly

tomocrafter:
I noticed that specifying `default_value = "human"` in `#[arg]` directive does not rely on Default trait for Format enum which results in always defaults to human even on GitHub Actions.
So I replaced it to `default_value_t` which calls `Default::default` and sets default value depends on GitHub Actions or not properly.
I also implemented Display for Format enum since `default_value_t` requires `impl ToString`, but clippy suggests `impl Display` instead of `impl ToString`.

### DIFF
--- a/crates/cli/src/commands.rs
+++ b/crates/cli/src/commands.rs
@@ -1,7 +1,5 @@
-use std::{
-    fmt::{write, Display},
-    path::PathBuf,
-};
+use std::path::PathBuf;
+use strum_macros::Display;
 
 use clap::{Parser, Subcommand, ValueEnum};
 
@@ -48,7 +46,8 @@ pub(crate) struct FixArgs {
     pub format: Format,
 }
 
-#[derive(Debug, Clone, Copy, ValueEnum)]
+#[derive(Debug, Clone, Copy, ValueEnum, Display)]
+#[strum(serialize_all = "kebab-case")]
 pub(crate) enum Format {
     Human,
     GithubAnnotationNative,
@@ -60,15 +59,6 @@ impl Default for Format {
             Format::GithubAnnotationNative
         } else {
             Format::Human
-        }
-    }
-}
-
-impl Display for Format {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            Format::Human => f.write_str("human"),
-            Format::GithubAnnotationNative => f.write_str("github-annotation-native"),
         }
     }
 }
--- a/crates/cli/src/commands.rs
+++ b/crates/cli/src/commands.rs
@@ -1,4 +1,7 @@
-use std::path::PathBuf;
+use std::{
+    fmt::{write, Display},
+    path::PathBuf,
+};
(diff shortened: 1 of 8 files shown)


crates/cli/src/commands.rs
```diff
@@ -60,3 +63,12 @@ impl Default for Format {
...
 }
+
+impl Display for Format {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Format::Human => f.write_str("human"),
+            Format::GithubAnnotationNative => f.write_str("github-annotation-native"),
+        }
+    }
+}
```
benfdking: nit: I wonder if we ca generate these automatically with `strum` and something like below?

```
use strum_macros::Display;

#[derive(Display)]
#[strum(serialize_all = "kebab-case")]
enum Status {
    InProgress,
    NotStarted,
    UnderReview,
}
```

benfdking (changes requested):
Thank you so much for debugging this! Big win for us. 

From my perspective, there are two little things: 
1. That nit which I think is a nice little win to avoid typos
2. I think it might be worth writing a `ui-test` for this as it's already once slipped through our checks.

Happy to do this oursleves with an additional commit or wait for you to do. Let us know and thank you again!

crates/cli/src/commands.rs
tomocrafter: Thank you for interesting to merge this PR, and of course this great library.

> That nit which I think is a nice little win to avoid typos

I usually avoid using library (crate) for small thing like this because:
1. you need to maintain the crate version
2. you need to go to that crate's repository to read code when you want to know what it does inside. feels it's overkill for what I want to do.

But you don't hesitate to add dependency to library, then I will add it and rewrite as you suggested (I also think that is clean tbh, I just worried to adding dependency).

> I think it might be worth writing a ui-test for this as it's already once slipped through our checks.

I would love to add test for it! but it might take some time like couple days, so if you want to merge this PR asap then 
 I am sorry but please consider adding it yourself.

nikolaevx: We’re already using this crate (just not in the one we’re currently working on), so it would be great to make it a workspace dependency.

benfdking: > I usually avoid using library (crate) for small thing like this ...

@tomocrafter, amen to that!

As @gvozdvmozgu pointed out, we already use it pretty heavily like [here](https://github.com/quarylabs/sqruff/blob/fb6adeb5b9544b76415bcbf949240882cb889396/crates/lib-core/src/dialects/syntax.rs#L9) so that's the only reason. 

I'll make the change, write the tests, and tack it on to this branch! Thanks again @tomocrafter!

@gvozdvmozgu good point on also making this a workspace dependency.

tomocrafter: Oh I completely missed that, then there is no reason to not use it!
Thank you for the support @benfdking adnd @gvozdvmozgu !

benfdking (approved):

PR #1033, opened 2024-12-04
refactor: make github native a fomatter

benfdking:


crates/cli/tests/github/test_fail_whitespace_before_comma.sql
```diff
@@ -0,0 +1 @@
+SELECT 1 ,4
```
benfdking, line 1: It's not outputting anything which isn't right/

PR #1038, opened 2024-12-06
fix: index out-of-bounds issue in the LT05 rule when trying to remove elements from a sequence

bradleyess:
When running `sqruff` on a DBT model in Snowflake dialect, the `LT05` rule was panicking. This only removes the indexes when they exist, otherwise prints.

**Error:**
```sh
thread 'main' panicked at crates/lib/src/rules/layout/lt05.rs:166:21:
removal index (is 14) should be < len (is 10)
```

### DIFF
--- a/crates/lib/src/rules/layout/lt05.rs
+++ b/crates/lib/src/rules/layout/lt05.rs
@@ -163,7 +163,15 @@ FROM my_table"#
         }
 
         for idx in to_remove {
-            results.remove(idx);
+            if idx < results.len() {
+                results.remove(idx);
+            } else {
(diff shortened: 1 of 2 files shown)


benfdking:
Hey @bradleyess, 

Thank you so much for your contribution. I am slightly worried about this in its current shape. We don't currently support the Jinja template and so I am not sure your test proves the fix: I think we skip the test for templates we don't have yet. 

I am looking at the Jinja templater right now, so I can look into this after that having finished it. 

If you can prove that this test fails without any jinja that could be really helpful.

PR #1097, opened 2024-12-25
fix: maturin build by adding version

benfdking:


pyproject.toml
```diff
@@ -10,7 +10,7 @@ dev = [
     "pytest>=8.0.0",
     "pytest-cov>=6.0.0",
     "ruff>=0.8.0",
```
nikolaevx, line 12: Do we really need ruff?

benfdking: Sorry, I don't quite understand the question; we don't need ruff. 

It's just being used as a linter/formatter for the few python files, is there any particular reason to not use it or have it?

nikolaevx: Oh, my mistake, I thought it was in the regular dependencies, not in dev.

benfdking: No worries! Thanks for looking!

PR #1156, opened 2025-01-01
fix: respect specified config filename

### LINKED ISSUE #1131: [Bug]: rule AL01 is not applied
### What happened?

Using sqruff 0.21.12 on Linux, when I run

`echo 'select foo.bar from table1 foo' | sqruff fix -`

the result is unchanged:

```
select foo.bar from table1 foo
```

Even though we have explicit table aliasing on in the default config. What I would expect is what I get from sqlfluff:

```
echo 'select foo.bar from table1 foo' | sqlfluff fix --dialect ansi -
select foo.bar from table1 as foo  
```

It correctly adds the `as`.

As far as I can tell AL02 is working just fine (explicit column aliasing). It is just AL01 that is failing. The default config looks good for AL01, and I have even tried specifying a config with only:

```ini
[sqlfluff]
rules = AL01
```

and it still fails.

I am happy to help, and will keep looking into this.

### Version

0.21.12

### Checked in Playground

- [X] I have checked this bug in the [playground](https://playground.quary.dev/?secondary=Format).

bowmanjd:
Previously, the `load_config_at_path` function removed the base filename and searched for filenames listed in `filename_options`, such as .sqruff and .sqlfluff. This fix changes the logic so that the filename_options list is only used if the file path is an existing directory. If instead the file path is an existing file, then simply load the config from the file.

fixes #1131. I recognize the presenting issue was already worked around by commit c69a0c81252f71ea1 but it still seems this should be fixed in `load_config_at_path`, yes?

### DIFF
--- a/crates/lib/src/core/config.rs
+++ b/crates/lib/src/core/config.rs
@@ -352,19 +352,18 @@ impl ConfigLoader {
             ".sqruff", /* "pyproject.toml" */
         ];
 
-        let path = if path.is_dir() {
-            path
-        } else {
-            path.parent().unwrap()
-        };
         let mut configs = AHashMap::new();
 
-        for fname in filename_options {
-            let path = path.join(fname);
-            if path.exists() {
-                self.load_config_file(path, &mut configs);
+        if path.is_dir() {
+            for fname in filename_options {
+                let path = path.join(fname);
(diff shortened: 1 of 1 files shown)


benfdking:
Thanks for the contribution!

PR #1186, opened 2025-01-12
feat: parse postgres trigram operators

### LINKED ISSUE #1161: Could you please have the errors returned implement Display and Error?
### What would you like to see?

It would make simpler to work with sqruff on codebases that expect returned errors to implement std::error::Error.

### Checked issues

- [X] I have checked this request in the [issues](https://github.com/quarylabs/sqruff/issues).

lu-zero:
See https://www.postgresql.org/docs/current/pgtrgm.html

Fixes #1161


lu-zero:
It is missing tests for now.

gvozdvmozgu:
Let me know if you need any help with it!

lu-zero:
It seems to not be enough, so if you can spot what I'm missing would be great :)

crates/lib-dialects/src/postgres.rs
```diff
@@ -58,6 +58,11 @@ pub fn raw_dialect() -> Dialect {
             r#"->>|#>>|->|#>|@>|<@|\?\||\?|\?&|#-"#,
             SyntaxKind::JsonOperator
         ),
+        Matcher::regex(
+            "trgm_operator",
+            r#"%|<%|%>|<<%|%>>|<->|<<->|<->>|<<<->|<->>>"#,
```
nikolaevx:
```suggestion
            r#"(<<<->|<->>>|<<->|<->>|<->|<<%|%>>|%>|<%|%)"#,
```

lu-zero:
Thank you :)

gvozdvmozgu (approved):

benfdking:
Sorry I've been a little slow to get to this change.

I think it's great and thanks for the contribution. The thing I am slightly worried about is that unlike sqlfluff we shouldn't make extensions part of the default dialect. 

I think it should add the concept of modifiers, I'll have a look at doing this next.

benfdking:
This is now merged, thanks for the original contribution.

PR #1216, opened 2025-01-17
refactor: replace sptr polyfill with std equivalent

gvozdvmozgu:


benfdking:
@gvozdvmozgu look at the benchmark ;)

PR #1218, opened 2025-01-17
docs: Added Redshift to supported dialects

duvenagep:


benfdking:
Thanks for the contribution! I just added a copy we have to do 🙏

PR #1234, opened 2025-01-20
feat: Add verbosity config to OutputStreamFormatter

duvenagep:
The original implementation initialises the OutputStreamFormatter with the 'verbosity = 0`. However, the parameter `verbose = (0 - 2)` from the config does not override this value.

How similar to the original SQLFluff implementation do you want to keep scruff? There are a few additional parameters not yet set, which I would be happy to implement.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -110,6 +110,7 @@ pub(crate) fn linter(config: FluffConfig, format: Format, collect_parse_errors: 
             let formatter = OutputStreamFormatter::new(
                 output_stream,
                 config.get("nocolor", "core").as_bool().unwrap_or_default(),
+                config.get("verbose", "core").as_int().unwrap_or_default(),
             );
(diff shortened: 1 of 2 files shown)


benfdking:
In terms of your question on how similar, we want the sqruff to be to sqlfluff, I think there's quite a bit we've kept the same  dialects, rules, etc. I think we will eventually start departing from sqlfluff especially when it comes to config.

benfdking (approved):

duvenagep:
@benfdking No Problem! Re-writing SQLFluff to rust was something I wanted to undertake for perf reasons as well, glad someone else also had the same idea. I think there a few `rust` specific improvements we could implement but just wanted to gauge your design philosophy.

I think a big area could be error handling. Thiserror for the lib & related crates and Anyhow for the CLI. Happy to keep contributing PRs. 

Great job btw!

PR #1239, opened 2025-01-23
refactor: simplify point extraction in `ReflowSequence` iterator

gvozdvmozgu:


crates/lib/src/utils/reflow/sequence.rs
```diff
@@ -440,28 +440,19 @@ impl<'a> ReflowSequence<'a> {
...
-                    (elem, pre, post).into()
-                }
-            } else {
-                None
+            let point = elem.as_point()?;
+            let mut pre = None;
+            let mut post = None;
+
+            if idx > 0 {
+                pre = Some(self.elements[idx - 1].as_block().unwrap());
```
benfdking, line 448: Are we sure we want to do this? Introducing unwraps always feels kind of wrong?

nikolaevx: If there's a panic here, then we have a bug and we've created the reflow IR incorrectly

benfdking: 👍

PR #1336, opened 2025-02-16
feat(sqruff): Introduce Rules & RulesFormatter to CLI

duvenagep:
This PR introduces the **Rules** command that lists the currently active rules similar in nature to `sqlfluff rules`. 
A `--nocolor` flag is also included to print to stdout without terminal colors.

Below are examples of the added functionality to the CLI:
Screenshot 2025-02-16 at 17 09 30
Screenshot 2025-02-16 at 17 10 04

The results as follows:

| Color | Black&White |
|-------------|--------------|
| Screenshot 2023-12-31 at 18 50 50     | Screenshot 2023-12-31 at 18 52 52      |


crates/cli/src/commands.rs
```diff
@@ -39,6 +39,8 @@ pub(crate) enum Commands {
         about = "Print information about sqruff and the current environment"
     )]
     Info,
+    #[command(name = "rules", about = "Show the current rules in use")]
```
benfdking: Could we just change this messaging slightly to something more akin to "rules available"? When I initially read it, it read as "rules that are turned on," which also isn't a bad thing to show, but this doesn't do.

crates/cli/src/commands.rs
```diff
@@ -60,6 +62,13 @@ pub(crate) struct FixArgs {
     pub format: Format,
 }
 
+#[derive(Debug, Parser)]
+pub(crate) struct RuleArgs {
+    /// No color - output will be without ANSI color codes.
+    #[arg(long)]
```
benfdking: nit (more a thought/opinion than something problematic): 

I feel like we should either detect it from the console or just print by color as a non-color terminal would just remove the colour. I don't know; it just feels like a flag we could remove and do implicitly

benfdking (changes requested):
I think just change the message and also you'll need to run 

```
cargo run --feature codegen-docs
```

to fix the doc thing

crates/cli/src/commands.rs
duvenagep: @benfdking Thanks for the suggestion. I tend to agree with you. 

My starting point was to have feature parity with `sqlfluff` as can be seen below:
Screenshot 2025-02-18 at 09 40 36

`ruff` has a slightly more descriptive message which I think is more in line with the actual functionality:
Screenshot 2025-02-18 at 09 44 55


Perhaps a combination of the 2 is the most appropriate, something like:

`Explain the available rules`

and in a future version of this feature I will add the ability to specify a rule then it can be updated to:

`Explain the available rules (or rule)`

Let me know what you think before I update the feature?

crates/cli/src/commands.rs
duvenagep: @benfdking again here my starting point was feature parity with `sqlfulff`, as can be seen below:
Screenshot 2025-02-18 at 09 41 15

This `nocolor` flag can be specified in the cfg file thus seems reasonable to specify it directly from the CLI. However, I did see that you have a function called `fn should_produce_plain_output(nocolor: bool) -> bool `. I can re-use this to auto-detect the colorized output.

I am currently duplicating code in the form of:
- `colorize_helper`
- `colorize`

plus now I will need `fn should_produce_plain_output(nocolor: bool) -> bool `. May I suggest that I abstract these functions into a `utils.rs` within the formatters crate? That way we can re-use the same code in many places thus ensuring consistency?

duvenagep: Thanks for the reference to:
`cargo run --feature codegen-docs`

I will update all code after clarity on the above suggestions

crates/cli/src/commands.rs
benfdking: Ah that makes sense. 

The ruff suggestion sounds good! And then we can add specific rules.

crates/cli/src/commands.rs
benfdking: Yep centralising makes sense! 

Thanks again for the contribution.

benfdking (approved):
Amazing work 🙏  Thanks for the contribution. 

My only one addition would be to add a "UI" test that checks in the output to make sure we track changes as CLI outputs are so hard to test but not letting that stop this so.

duvenagep:
@benfdking Good suggestion.

I am planning to add an additional feature, which allows users to specify a rule `code` in the rules CLI. I will have a look at the testing setup and add the necessary tests. I will reach out if needed.

My pleasure, can't wait to have `sqruff` be the de facto SQL Formatter & Linter. It's performance is amazing!

Issue #1299, opened 2025-02-06
[Bug]: Missing Linux Binaries since 0.25.2


Joelius300:
### What happened?

For 0.24.3 and before, there were binary versions for Linux available on PyPi (see [here](https://pypi.org/project/sqruff/0.24.3/#files)).

Since 0.25.2, this is not the case anymore (see [here](https://pypi.org/project/sqruff/0.25.2/#files)). This is breaking various workflows for us.

Is there a reason you stopped publishing Linux binaries? If possible, please bring them back.

### Version

0.25.2

### Checked in Playground

- (N/A) I have checked this bug in the [playground](https://playground.quary.dev/?secondary=Format).

benfdking:
Sorry about that. We will reintroduce it. A few issues caused this for us.

benfdking:
We've repushed but I looked and unfortunately our changes are making the releases python specific because of the apis we're calling, if you could let us know the details we can try to get that quicker to you. 

At the moment I have readded x86 and 3.12

Joelius300:
Thanks for the fast response! ❤ 

We're on RHEL9.
- For 0.24.3, we consumed `sqruff-0.24.3-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`.
- For 0.25.2, it tries to pull the sources and compile them with maturin, which doesn't work and isn't desirable in those environments.
- For 0.25.3, it still tries to compile from source, just like 0.25.2 because none of the binary distributions are compatible.

On my Manjaro machine, it consumes `sqruff-0.24.3-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl` and `sqruff-0.25.3-cp312-cp312-manylinux_2_39_x86_64.whl`, both work fine without compilation. But apparently, it's not enough for RHEL9.

benfdking:
Ok let me look into it and see if I can recreate and improve the situation.

Joelius300:
FYI the maturin build fails is because it is pulling `rustc 0.82.0` as build dependency, but it seems that "`sqruff@0.25.3` requires `rustc 1.84.1`". Maybe this could be fixed and then local compilation from source would work.

Still, for us it's more important to have pre-built binaries.

Joelius300:
Just tested with v0.25.4 and the problem persists.

benfdking:
Yep v0.25.4 mainly addressed other issues. Sorry about this persisting.

concur1:
[...]

PR #1440, opened 2025-03-22
fix: python whl builds for some platforms.

concur1:
This is a partial fix for #1299 , I think it should fix x86 Linux, Windows and aarch64 MacOS.

Updated CI to use python extension module to call binaries as suggested by the Maturin maintainer. Much of the code in this commit was copied from the Ruff project (MIT License) and adapted.

- Updated to use python extension module as suggested by the maturin maintainer.
- Copied most of the code for the python extension module from the ruff project.
- This code now has some minor tests that run after the whl's have been created.

### Problems:
- The x86 macos build fails when running on a aarch64 runner due to a cross compilation error. Error message here: https://github.com/quarylabs/sqruff/actions/runs/14007557827/job/39223363737
trying to get it to run on a x86 runner doesn't seem to work, the macos-12 and macos-11 runners don't seem to start. **Therefore I have commented them out for the time being.**


benfdking:
This is awesome. Let me have a look at it today!

concur1:
> This is awesome. Let me have a look at it today!

Thanks! Let me know if you have any questions!

concur1:
Hi @benfdking, thanks for merging!

It looks like the release didn't work properly/find any files and there is no new release on pypi. I think this might be because of changing line 404 in `python-ci.yml` from:
```yaml
    args: --non-interactive --skip-existing wheels-*/*
```
to:
```yaml          
    args: --non-interactive --skip-existing artifacts-*/
```

I believe we are still using the 'wheels-' prefix for the wheels we are uploading as artifacts so we still want to use the `wheels` prefix in the release args on line 404.

PR #1486, opened 2025-04-03
chore: update Rust version to 1.86

gvozdvmozgu:


gvozdvmozgu:
@concur, why does sdist use an older version of the Rust compiler?

Issue #1028, opened 2024-12-02
Support multiple statements in one file


earthboundkid:
### What would you like to see?

If I have more than one statement in a .sql file, sqruff only lints the first statement.

### Checked issues

- [ ] I have checked this request in the [issues](https://github.com/quarylabs/sqruff/issues).

gvozdvmozgu:
Could you please provide an example of the request and configuration, if available? The issue is likely with the parser.

earthboundkid:
.sqruff

```
[sqruff]
dialect = sqlite
rules = all

[sqruff:indentation]
indent_unit = space
tab_space_size = 2
indented_joins = True

[sqlfluff:rules:capitalisation.keywords]
# Keywords
capitalisation_policy = upper
# Comma separated list of words to ignore for this rule
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:capitalisation.identifiers]
# Unquoted identifiers
extended_capitalisation_policy = lower
# Comma separated list of words to ignore for this rule
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:capitalisation.functions]
# Function names
extended_capitalisation_policy = lower
# Comma separated list of words to ignore for this rule
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:capitalisation.literals]
# Null & Boolean Literals
capitalisation_policy = upper
```

comment.sql

```sql
-- name: CreateComment :one
INSERT INTO
comment (
  name, contact, subject, cc, message, ip, user_agent, referrer, host_page
)
VALUES
(
  ?, ?, ?, ?, ?, ?, ?, ?, ?
)
RETURNING * ;

-- name: ListComments :many
SELECT *
FROM comment
ORDER BY modified_at DESC
LIMIT ?
OFFSET ? ;
```

If I swap the statements, it unindents the insert columns and question marks.

benfdking:
I'm sorry, but I've renamed this because the issue is not quite what you've said. 

The error is two-fold: 

1. We don't parse the following syntaxes yet `RETURNING *`.
2. Your statements are not valid SQL that's why in parts it's struggling. `?` isn't valid SQL. You're looking for the [placeholder templater](https://github.com/quarylabs/sqruff/blob/main/docs/templaters.md#placeholder)

```sql
-- name: CreateComment :one
INSERT INTO
comment (
  name, contact, subject, cc, message, ip, user_agent, referrer, host_page
)
VALUES
(
[...]

PR #1490, opened 2025-04-03
fix: add ability to parse returning in sqlite delete update

karimkhaleel:
Queries including the `RETURNING` clause in `DELETE` and `UPDATE` statements and what comes after them were not being properly parsed in the sqlite dialect. 

I am not sure if this is the best way to go about fixing this but this patch got both the `DELETE` and `UPDATE` queries with a `RETURNING` clause formatting properly. 

Sqlite supports [DELETE](https://sqlite.org/lang_delete.html), [INSERT](https://sqlite.org/lang_insert.html), and [UPDATE](https://sqlite.org/lang_update.html) statements. https://sqlite.org/lang_returning.html

Related to https://github.com/quarylabs/sqruff/issues/1028

### DIFF
--- a/crates/lib-dialects/src/sqlite.rs
+++ b/crates/lib-dialects/src/sqlite.rs
@@ -449,6 +449,41 @@ pub fn raw_dialect() -> Dialect {
             .to_matchable()
             .into(),
         ),
+        (
+            "DeleteStatementSegment".into(),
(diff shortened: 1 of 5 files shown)


benfdking (approved):
Thanks for the contribution!

karimkhaleel:
Just signed my commit

PR #1500, opened 2025-04-09
feat: replace `rayon` with `par-iter` for parallel iteration

gvozdvmozgu:


gvozdvmozgu:
Okay, that's a significant regression.

PR #1512, opened 2025-04-13
fix: try compiling x86 without jemallocator

benfdking:


gvozdvmozgu:
It appears that the performance dropped from 52.162579 ms to 75.026102 ms.

benfdking:
Yep, I want to reintroduce it but it wasn't compiling.

PR #1534, opened 2025-04-23
feat: fix python working with pip install

benfdking:


benfdking:
@concur1, would you be able to test this change before I merge it? I want to make sure I don't break anything for you.

concur1:
Thanks for checking @benfdking, I have tested for x86 linux and arm macos and was able to install the wheels for both so as far as I can tell it the python package should be fine.

PR #1536, opened 2025-04-25
refactor: remove legacy severity flags from SQLBaseError and streamline formatters

gvozdvmozgu:


benfdking:
I am not sure this is the right approach:
1. is the complexity worth it compared to an enum?
2. Some of these properties like whether something is fixable or a warning is orthogonal and so be separate fields?

gvozdvmozgu:
Hmm, looking at the code now, I think we can just introduce a `Severity` enum with `Fatal` and `Warning` variants (we never actually use `IGNORE`, so we can drop it). `fixable` can be a separate `bool` field, which won’t affect the struct’s size.

benfdking:
> Hmm, looking at the code now, I think we can just introduce a `Severity` enum with `Fatal` and `Warning` variants (we never actually use `IGNORE`, so we can drop it). `fixable` can be a separate `bool` field, which won’t affect the struct’s size.

Sounds great!

PR #1558, opened 2025-04-29
refactor(rules): migrate all `Rule` impls to `define_rule!` macro

gvozdvmozgu:


gvozdvmozgu:
@benfdking it seems much simpler, and in the future it will be easier if we want to remove the `Rule` trait.

benfdking:
Hey 👋 

This to me feels somewhat simpler and more complex in the same breath: Cleaner as the rule code itself looker simpler, less approachable because of that macro. 

I honestly don't know which is better, so happy to go with your recommendation. 

As I see it at the moment, there's a neat win in having rule descriptions also apply in Rust docs. Curious to understand though why you think we can drop the trait/why we would want that.

gvozdvmozgu:
By removing the `Rule` trait, we avoid cloning the rule when generating linter warnings (i.e., cloning the entire struct, not just a pointer, which can be costly if the rule carries configuration) and can finally eliminate the `dyn_clone` dependency.

gvozdvmozgu:
However, this will probably also require modifying the configurator, as suggested in an issue, to consolidate everything into a single large `Config` struct.

benfdking:
> By removing the Rule trait, we avoid cloning the rule when generating linter warnings (i.e., cloning the entire struct, not just a pointer, which can be costly if the rule carries configuration) and can finally eliminate the dyn_clone dependency.

Happy to give it a go!

PR #1559, opened 2025-04-29
refactor(rules): extract crawl logic into standalone functions

gvozdvmozgu:


benfdking:
Any understanding of why everything got slower?

gvozdvmozgu:
Since merge #1554, performance has significantly decreased.
Execution took: 332.133794ms on that PR.

benfdking:
Yep I had to do that painful fix to get the python stuff working.

gvozdvmozgu:
I don't have these issues locally, so it's possible that GitHub Actions is just under heavy load?

gvozdvmozgu:
It seems that #1561 fixes the problem, and the numbers are more convincing again.

PR #1564, opened 2025-04-30
feat: accept sqlc style placeholders

bamorim:
I'm not a Rust expert, but opening this as a conversation starter. Not sure if this is all it is needed.

So, I've been using sqlc in a golang project and I tried to use named variables as described [here](https://docs.sqlc.dev/en/latest/howto/named_parameters.html) and it caused problems with sqruff formatting.

This is my attempt at supporting that.

EDIT: Tested with the following config in my project and it worked perfectly

```
[sqruff]
templater = placeholder

param_style = at
```

### DIFF
--- a/crates/lib/src/templaters/placeholder.rs
+++ b/crates/lib/src/templaters/placeholder.rs
@@ -191,6 +191,9 @@ A few common styles are supported:
 
  -- apache_camel
  WHERE bla = :#${qwe}
+
+ -- at
+ WHERE bla = @my_name
 ```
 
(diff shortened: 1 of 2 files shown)


benfdking:
Hey @bamorim, happy to add this! Thanks for the contribution! 

Would you just be able to add it to the documentation for the templater, and then also generate the docs with 

```bash
cargo run --features codegen-docs
```

Then very happy to ship it!

benfdking:
That's being released right now but there's an issue with the pipeline. It still gets pushed to pypi so if you pip install from there it will have this change.

PR #1572, opened 2025-05-02
ci: use dist as a distribution engine instead

benfdking:


gvozdvmozgu:
It looks like the original cargo-dist is no longer maintained, but there’s a fork available at https://github.com/astral-sh/cargo-dist.

PR #1675, opened 2025-06-23
fix: Implement LookaheadExclude to resolve T-SQL WITH keyword parsing issue

### LINKED ISSUE #1673: T-SQL: WITH keyword incorrectly parsed as alias instead of table hint
## Description

In T-SQL, the `WITH` keyword following a table reference should be parsed as the beginning of a table hint (e.g., `WITH (NOLOCK)`), but it's currently being incorrectly parsed as an alias. This causes linting rule AL05 to report false positives about unused aliases.

## Current Behavior

When parsing this T-SQL query:
```sql
SELECT * FROM Users WITH (NOLOCK)
```

The parser incorrectly interprets `WITH` as an alias:
```yaml
table_expression:
  - table_reference:
    - naked_identifier: Users
- alias_expression:
  - naked_identifier: WITH  # ❌ Wrong - should be part of table hint
- bracketed:
  - start_bracket: (
  - naked_identifier: NOLOCK
  - end_bracket: )
```

This triggers a false positive from rule AL05:
```
L:   1 | P:  15 |  AL05 | Alias 'WITH' is never used in SELECT statement.
```

## Expected Behavior

The `WITH (NOLOCK)` should be parsed as a table hint:
```yaml
table_expression:
  - table_reference:
    - naked_identifier: Users
  - keyword: WITH
  - bracketed:
    - start_bracket: (
    - keyword: NOLOCK
    - end_bracket: )
```

### Root Cause: Fundamental Parser Architecture Limitations

This is a **fundamental architectural limitation** in the parser/linter system. The issue stems from architectural limitations in the core parser that prevent any dialect-level fix:

#### 1. Greedy Left-to-Right Matching Strategy
**Location**: `/crates/lib-core/src/parser/grammar/sequence.rs` (lines 119-249)

The parser uses greedy sequential matching where elements are processed in order with no backtracking. Once an optional element matches, it consumes those tokens permanently.

#### 2. Problematic Sequence Order in FromExpressionElementSegment
**Location**: `/crates/lib-dialects/src/ansi.rs` (lines 4386-4411)

```rust
Sequence::new(vec_of_erased![
    Ref::new("PreTableFunctionKeywordsGrammar").optional(),
    optionally_bracketed(vec_of_erased![Ref::new("TableExpressionSegment")]),
    Ref::new("AliasExpressionSegment")     // <-- This comes BEFORE table hints!
        .exclude(/* ... */)
        .optional(),
    // ... other patterns ...
    Ref::new("PostTableExpressionGrammar").optional()  // <-- Table hints come LAST!
])
```

**Critical Issue**: `AliasExpressionSegment` appears before `PostTableExpressionGrammar` (which contains table hints), giving aliases first opportunity to claim WITH tokens.

#### 3. Limited Exclude Mechanism - No Lookahead Capability
**Location**: `/crates/lib-core/src/parser/grammar.rs` (lines 134-151)

The current exclude mechanism **cannot perform lookahead**:
```rust
if exclude.match_segments(segments, idx, this).is_ok_and(|match_result| match_result.has_match()) {
    return Some(MatchResult::empty_at(idx));
}
```

**Limitation**: Exclude patterns only match at the exact current position (`idx`). There's no ability to check if "WITH" is followed by "(" to distinguish table hints from aliases.

## Comprehensive Testing - 12 Approaches Attempted

I systematically tested 12 different approaches to resolve this issue:

1. ✅ **Verified base components work correctly**:
   - WITH is in T-SQL reserved keywords (`tsql_keywords.rs:196`)
   - TableHintSegment definition is correct and works in isolation
   - PostTableExpressionGrammar includes TableHintSegment correctly
   - WITH is correctly tokenized as keyword (not identifier)

2. ❌ **All attempted fixes failed**:
   - Override FromExpressionElementSegment with explicit patterns
   - Add `exclude(Ref::keyword("WITH"))` to alias patterns
   - Use lookahead exclusions for "WITH(" pattern
   - Regex-based table hint parsing
   - Enhanced exclude lists with table hint exclusions
   - Pattern prioritization with `one_of()`
   - Direct TableHintSegment placement bypassing PostTableExpressionGrammar
   - Custom lexer matchers for table hint recognition
   - Corrected exclude() syntax with config closures
   - Override NakedIdentifierSegment with forced WITH exclusion
   - Custom table reference parser with table hint precedence
   - **Enhanced anti-template exclusions in custom AliasExpressionSegment** (latest attempt)

### Why All Dialect-Level Attempts Failed

1. **Anti-template mechanisms** work at the RegexParser level but don't prevent higher-level grammar matching
2. **Exclude patterns** cannot lookahead to distinguish "WITH (" from "WITH identifier"  
3. **Pattern reordering** within T-SQL can't override ANSI's FromExpressionElementSegment sequence
4. **Custom parsers** still operate within the greedy matching framework

## Proposed Solution: Implement LookaheadExclude in Core Parser

The optimal solution is to **extend the core exclude mechanism with lookahead capability**.

#### 1. Create New LookaheadExclude Matcher
**New file**: `/crates/lib-core/src/parser/lookahead.rs`

```rust
pub struct LookaheadExclude {
    first_token: String,      // e.g., "WITH"
    lookahead_token: String,  // e.g., "("
}

impl MatchableTrait for LookaheadExclude {
    fn match_segments(&self, segments: &[ErasedSegment], idx: u32, parse_context: &mut ParseContext) 
        -> Result<MatchResult, SQLParseError> {
        // Check if current token matches first pattern
        if segments[idx as usize].raw().to_uppercase() == self.first_token {
            // Look ahead for second token (skipping whitespace)
            let next_idx = skip_start_index_forward_to_code(segments, idx + 1, segments.len() as u32);
            if next_idx < segments.len() as u32 {
                if segments[next_idx as usize].raw() == self.lookahead_token {
                    // Match found - exclude this position
                    return Ok(MatchResult::from_span(idx, idx + 1));
                }
            }
        }
        // No match - don't exclude
        Ok(MatchResult::empty_at(idx))
    }
}
```

#### 2. Integrate with ANSI Dialect
**Modify**: `/crates/lib-dialects/src/ansi.rs`

```rust
Ref::new("AliasExpressionSegment")
    .exclude(one_of(vec_of_erased![
        Ref::new("FromClauseTerminatorGrammar"),
        Ref::new("SamplingExpressionSegment"),
        Ref::new("JoinLikeClauseGrammar"),
        // NEW: Exclude WITH when followed by opening parenthesis
        LookaheadExclude::new("WITH", "(").to_matchable()
    ]))
    .optional(),
```

### Why This Solution is Optimal

1. **Surgical Precision**: Only affects the specific problematic pattern without disrupting other parsing
2. **Minimal Architecture Impact**: Extends existing exclude infrastructure rather than redesigning it
3. **Cross-Dialect Benefits**: All dialects inheriting from ANSI automatically get the fix
4. **Future-Proof**: Creates reusable infrastructure for similar parsing conflicts
5. **Performance Efficient**: Lookahead only occurs during exclude evaluation, not general parsing

## Impact

- False positive linting errors for valid T-SQL queries using table hints
- Cannot properly lint T-SQL code that uses `WITH (NOLOCK)` or similar hints
- Affects JOIN clauses as well: `JOIN Users WITH (NOLOCK) ON ...`

## Test Cases

Multiple test files demonstrate this issue:
- `al05_alias_in_where_clause.yml` - Shows WITH parsed as alias in JOIN
- `al05_exact_issue.yml` - Minimal reproduction case
- `apply_clause.yml` - Shows issue in OUTER APPLY context

## Workaround

Currently, users can disable AL05 rule for T-SQL files with table hints until this core parser enhancement is implemented.

## Additional Context

This issue was discovered while implementing T-SQL dialect support in PR #1664. The problem is documented in detail in `crates/lib-dialects/src/tsql.rs` lines 23-53, including all attempted fixes and why they failed.

The core challenge is that the parser's pattern matching and exclusion mechanisms don't properly handle this specific case where a reserved keyword immediately follows a table reference. This requires enhancing the core parser's exclude mechanism with lookahead capability rather than dialect-level fixes.

Fank:
## Summary

This PR implements a comprehensive fix for the T-SQL table hints parsing issue where `WITH (NOLOCK)` was incorrectly parsed as an alias instead of a table hint, triggering false AL05 linter errors.

Fixes #1673

## Problem

The issue stemmed from fundamental architectural limitations in the core parser:
- `AliasExpressionSegment` appears before `PostTableExpressionGrammar` in the parsing sequence
- The parser uses greedy left-to-right matching with no backtracking
- The exclude mechanism lacked lookahead capability to distinguish `WITH (` from `WITH identifier`

## Solution

This PR introduces a new `LookaheadExclude` matcher that extends the core parser's exclude mechanism with lookahead capability:

### Core Parser Changes
- **New `LookaheadExclude` matcher** (`lib-core/src/parser/lookahead.rs`)
  - Implements lookahead to check if a token is followed by a specific pattern
  - Case-insensitive matching for the first token
  - Skips whitespace between tokens using existing utilities
- **Integration with `MatchableTraitImpl`** enum
- **Module exports** in parser.rs

### ANSI Dialect Changes
- Added `LookaheadExclude::new("WITH", "(")` to `AliasExpressionSegment` exclude list
- This prevents WITH from being parsed as an alias when followed by an opening parenthesis
- All dialects inheriting from ANSI automatically benefit from this fix

### T-SQL Dialect Updates
- Comprehensive documentation of 12 attempted dialect-level approaches that failed
- Investigation notes for future reference
- Restored previously removed test files that demonstrate the issue

## Testing

The implementation has been tested with various SQL patterns:

✅ **T-SQL table hints now work correctly**:
```sql
SELECT * FROM Users WITH (NOLOCK)              -- No AL05 error
SELECT * FROM Orders WITH (READUNCOMMITTED)    -- No AL05 error
```

✅ **Other WITH patterns remain unaffected**:
```sql
WITH CTE AS (SELECT 1 AS id) SELECT * FROM CTE -- CTEs work correctly
SELECT * FROM Products p                        -- Regular aliases work
```

✅ **Cross-dialect compatibility maintained** - ANSI SQL and other dialects continue to work as expected

## Test Files Restored

This PR also restores T-SQL test files that were previously removed:
- `al05_alias_in_where_clause.sql/yml`
- `al05_exact_issue.sql/yml` 
- `tsql_alias_column_ref.sql/yml`
- `tsql_alias_simple.sql/yml`

These files demonstrate the exact issue being fixed and serve as regression tests.

## Impact

This fix resolves a long-standing issue that affected T-SQL users and establishes infrastructure for solving similar parsing conflicts in other SQL dialects. The solution is surgical and targeted, affecting only the specific problematic pattern without disrupting other parsing behavior.


Fank:
@benfdking should be ready for review.

crates/lib-core/src/parser/lookahead.rs
```diff
@@ -0,0 +1,89 @@
...
+/// A matcher that excludes patterns based on lookahead.
+///
+/// This is useful for cases where we need to exclude a token (like "WITH")
+/// only when it's followed by a specific pattern (like "(").
+#[derive(Debug, Clone, PartialEq)]
+pub struct LookaheadExclude {
+    /// The first token to match (e.g., "WITH")
+    first_token: String,
+    /// The lookahead token to check for (e.g., "(")
+    lookahead_token: String,
```
nikolaevx: Could we just use a `&'static str` here?

fank: for both?

nikolaevx: If that’s possible, then yes

fank: Good suggestion! I've updated the code to use `&'static str` for both fields. This avoids unnecessary allocations since these are always string literals. 

Changes made in commit 8483c707:
- Changed `first_token` and `lookahead_token` to `&'static str`
- Updated constructor signature 
- Switched from `to_uppercase()` to `eq_ignore_ascii_case()` for more efficient comparison

Thanks for the review!

fank: I am still not that familiar with rust, so If you notice anything just let me know.

crates/lib-core/src/parser/lookahead.rs
```diff
@@ -0,0 +1,89 @@
...
+        // Check if current token matches first pattern (case-insensitive)
+        let current_raw = segments[idx as usize].raw();
+        if current_raw.eq_ignore_ascii_case(self.first_token) {
+            // Look ahead for second token, skipping any whitespace
+            let next_idx =
+                skip_start_index_forward_to_code(segments, idx + 1, segments.len() as u32);
+
+            if next_idx < segments.len() as u32 {
+                let next_raw = segments[next_idx as usize].raw();
+                if next_raw == self.lookahead_token {
```
nikolaevx: You can also use eq_ignore_ascii_case here.

fank: Good catch! Updated in commit 288bb383. Now both token comparisons use `eq_ignore_ascii_case` for consistency.

fank: replaced

PR #1695, opened 2025-06-25
fix: Handle T-SQL APPLY clauses in JoinClauseSegment to prevent panic

### LINKED ISSUE #1691: Panic in AL05 rule when processing T-SQL APPLY clauses
## Description

The linter panics with `called Option::unwrap() on a None value` when processing T-SQL queries that contain `OUTER APPLY` or `CROSS APPLY` clauses. This affects multiple aliasing rules (AL04, AL05) and reference rules (RF01).

## Error Message

```
thread 'main' panicked at crates/lib-core/src/parser/segments/join.rs:15:14:
called `Option::unwrap()` on a `None` value
```

## Reproduction

Create a file `test.sql` with the following content:

```sql
SELECT
    COALESCE(Orders.TotalAmount, 0) AS OrderTotal
FROM
    Customers AS c WITH(NOLOCK)
    OUTER APPLY (
        SELECT
            TotalAmount = SUM(Amount)
        FROM
            Orders AS o WITH(NOLOCK)
        WHERE
            o.CustomerID = c.CustomerID
    ) AS Orders
WHERE
    c.CustomerID = 123
```

Run with T-SQL dialect:
```bash
echo "[sqruff]
dialect = tsql" > .sqruff

sqruff lint test.sql
```

## Root Cause

The issue occurs because:

1. T-SQL's `ApplyClauseSegment` is marked with `SyntaxKind::JoinClause`
2. The `JoinClauseSegment::eventual_aliases` function assumes all join clauses have a `FromExpressionElement` child
3. APPLY clauses have a different structure - the `FromExpressionElementSegment` is directly in the sequence, not nested
4. The `unwrap()` fails because `child()` returns `None` for APPLY clauses

## Stack Trace

```
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: sqruff_lib_core::parser::segments::join::JoinClauseSegment::eventual_aliases
   5: sqruff_lib_core::utils::analysis::select::get_select_statement_info
   6: <sqruff_lib::rules::aliasing::al05::RuleAL05 as sqruff_lib::core::rules::Rule>::eval
```

## Affected Components

- `crates/lib-core/src/parser/segments/join.rs` - Line 15
- Rules: AL04, AL05, RF01 (any rule that calls `get_select_statement_info`)

## Environment

- sqruff version: 0.26.8
- Dialect: tsql

## Workaround

Currently none - the panic prevents linting of any T-SQL file containing APPLY clauses.

## Suggested Fix

Update `JoinClauseSegment::eventual_aliases` to handle APPLY clauses by checking the join type and extracting aliases based on the appropriate structure for each join variant.

Fank:
## Description

This PR fixes the panic in AL05 rule when processing T-SQL queries that contain `OUTER APPLY` or `CROSS APPLY` clauses.

## The Issue

The linter was panicking with `called Option::unwrap() on a None value` when processing T-SQL APPLY clauses. This occurred because:

1. T-SQL's `ApplyClauseSegment` is marked with `SyntaxKind::JoinClause`
2. The `JoinClauseSegment::eventual_aliases` function assumed all join clauses have a `FromExpressionElement` child
3. APPLY clauses have a different structure - the `FromExpressionElementSegment` is directly in the sequence, not nested
4. The `unwrap()` failed because `child()` returns `None` for APPLY clauses

## The Fix

- Added detection for APPLY clauses by checking if any keyword child has "APPLY" as its raw value
- Used different logic to extract the `FromExpressionElement` based on whether it's an APPLY clause or regular join
- Made the extraction safe by handling the `Option` return value properly
- Fixed an additional bug where `join_clause.id() == join_clause.id()` was always comparing to itself

## Testing

Added comprehensive test cases:
- `apply_with_alias.sql/yml` - Tests OUTER APPLY with table aliases
- `cross_apply_with_function.sql/yml` - Tests CROSS APPLY with table-valued functions

The fix has been tested with the reproduction case from the issue and all tests pass.

Fixes #1691

### DIFF
--- a/crates/lib-core/src/parser/segments/join.rs
+++ b/crates/lib-core/src/parser/segments/join.rs
@@ -9,13 +9,25 @@ impl JoinClauseSegment {
     pub fn eventual_aliases(&self) -> Vec<(ErasedSegment, AliasInfo)> {
         let mut buff = Vec::new();
 
-        let from_expression = self
-            .0
-            .child(const { &SyntaxSet::new(&[SyntaxKind::FromExpressionElement]) })
-            .unwrap();
-        let alias = FromExpressionElementSegment(from_expression.clone()).eventual_alias();
+        // Check if this is an APPLY clause (CROSS APPLY or OUTER APPLY)
+        // APPLY clauses have a different structure where FromExpressionElement is in the sequence
+        let is_apply = self.0.children(const { &SyntaxSet::new(&[SyntaxKind::Keyword]) })
+            .any(|kw| kw.raw().to_uppercase() == "APPLY");
 
-        buff.push((from_expression.clone(), alias));
+        let from_expression = if is_apply {
+            // For APPLY clauses, find the FromExpressionElement in the sequence
+            self.0.children(const { &SyntaxSet::new(&[SyntaxKind::FromExpressionElement]) })
+                .next()
+                .cloned()
+        } else {
+            // For regular JOIN clauses, get the nested child
+            self.0.child(const { &SyntaxSet::new(&[SyntaxKind::FromExpressionElement]) })
+        };
+
+        if let Some(from_expr) = from_expression {
+            let alias = FromExpressionElementSegment(from_expr.clone()).eventual_alias();
+            buff.push((from_expr.clone(), alias));
+        }
 
         for join_clause in self.0.recursive_crawl(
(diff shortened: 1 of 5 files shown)


Fank:
So far after this I don't see any panics anymore in my codebase.
I think its the last one.

PR #1699, opened 2025-06-26
docs: updating contributing

NormallyGaussian:
This updates `contributing.md` to reflect what appears to be the correct way to set up your development environment for the first time. The existing instructions weren't complete as `cargo test` fails if the virtual environment isn't set up already.

I'm happy to make edits to reflect the correct setup. These steps worked for me.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Contributions are welcome! See [CONTRIBUTING.md](./CONTRIBUTING.md) for guidelin
 
 ## Running locally
 
-The project shouldn't require anything more than the `rust-toolchain.toml` version of Rust specified which you can set up with `rustup`. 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -4,10 +4,40 @@ Contributions are welcome! See [CONTRIBUTING.md](./CONTRIBUTING.md) for guidelin
 
 ## Running locally
 
-The project shouldn't require anything more than the `rust-toolchain.toml` version of Rust specified which you can set up with `rustup`. Once you have that the following runs the test suite. 
+The project shouldn't require anything more than the `rust-toolchain.toml` version of Rust specified which you can set up with `rustup`.
```
benfdking: The fact that you only need Rust isn't really accurate

```suggestion
The project requires Rust version specified in `rust-toolchain.toml`which you can set up with `rustup`, as well as a Python environment setup as shown below.
```

PR #1715, opened 2025-06-28
feat: back quotes in clickhouse

DamianMaslanka5:
Similar changes in sqlfluff https://github.com/sqlfluff/sqlfluff/pull/5457

I am not sure if I added `"BackQuotedIdentifierSegment".into(),` in the right place, but the yml looks good


benfdking:
Thank you so much for the contribution! I've used the backquote ticks tests form sqlfluff to make maintenance a little easier.

DamianMaslanka5:
@benfdking thanks for creating sqruff :)

It looks like you try to reuse tests from sqlfluff when possible, I have some questions related to maintenance

1. Do you also try to update the tests in sqruff when someone updates tests in sqlfluff?
2. Is the generated yml compatbile with sqlfluff?
3. Would it make sense to automate synchronizing [std_rule_cases](https://github.com/sqlfluff/sqlfluff/tree/main/test/fixtures/rules/std_rule_cases) tests?
    * For example, opening PR per yml file, after yml file is updated in sqlfluff
    * This could allow you to see how compatible sqruff is with sqlfluff and maybe make maintenance a bit easier

PR #1698, opened 2025-06-25
fix: Comprehensive T-SQL parser and rule fixes for operators, CREATE statements, and multi-line constructs

Fank:
## Summary
This PR fixes multiple T-SQL parser issues identified in #1685. The changes ensure that compound operators and Unicode literals are properly tokenized as single tokens, adds comprehensive support for CREATE statements, and enhances rule handling for T-SQL specific constructs.

### Lexer Fixes
1. **Compound Comparison Operators**: Added proper matchers for `>=`, `<=`, `!=`, and `<>` to parse as single tokens
2. **Unicode String Literals**: Added matcher for `N'string'` syntax to parse as single tokens  
3. **Compound Assignment Operators**: Added proper parsing for `+=`, `-=`, `*=`, `/=`, `%=`, `&=`, `|=`, `^=`

### Grammar Additions
1. **CREATE TABLE**: Full support including temp tables, IDENTITY columns, and AS SELECT syntax
2. **CREATE FUNCTION**: Support for both scalar and table-valued functions
3. **CREATE PROCEDURE**: Basic procedure creation support
4. **CREATE TYPE**: Table type definitions support
5. **IDENTITY Grammar**: Proper handling of IDENTITY(seed, increment) syntax

### CP01 Rule Enhancement
Fixed CP01 rule to capitalize T-SQL function names that are also keywords:
- Added `FunctionNameIdentifier` to CP01 crawl behavior
- Added logic to check if function names are also keywords in the dialect
- Only processes single-segment function names that are keywords
- Fixes issue where `CAST` and `COALESCE` weren't being capitalized in T-SQL

### CV01 Rule Enhancement
Enhanced CV01 rule to handle multi-line comparison operators:
- Added support for detecting comparison operators split across lines with comments/whitespace
- Handles cases like `WHERE 1 < -- comment \n > 2` and converts appropriately
- Added `eval_multiline_operator()` method to process split operators
- Updated crawl behavior to include Symbol and RawComparisonOperator tokens
- Maintains style consistency across single-line and multi-line operators

### Tests Re-enabled
Successfully re-enabled multiple tests that now pass with these fixes:
- `LT01-excessive.yml`: test_comparison_operator_pass
- `LT01-literals.yml`: Unicode literal tests (2 tests)  
- `LT01-operators.yml`: Compound assignment operator test
- `CP01.yml`: Function capitalization test (test_fail_select_lower_keyword_functions)
- `CV01.yml`: Multi-line comparison operator tests (2 tests) - **NEW**

[...]

PR #1717, opened 2025-06-28
chore: add test for keywords: upper and functions: lower

DamianMaslanka5:
I think https://github.com/quarylabs/sqruff/pull/1698 broke formatting for functions (I didn't verify that this exact PR brakes it), see the test in 7d9a438dfc68a5b9b4f6f04430bcbc78b8e0d330

I checked that this test works correctly on v0.27.0, but it doesn't work on main

cc: @Fank @benfdking


benfdking:
Thanks for this find!

PR #1738, opened 2025-07-02
[DRAFT] Test synchronizing std_rule_cases

DamianMaslanka5:
I created `update_test_cases.py` to copy test cases from sqlfluff repo (local folder) to sqruff.  
All content from sqlfluff is copied, only the `ignored` field in sqruff is preserved to allow keeping tests as ignored.

In `crates/lib/test.txt` is output of `cargo test --test rules -- --no-fail-fast > test.txt`, there are more `failed:` tests than I expected.

TODO:
1. Move test cases specific to sqruff into separate files/folder to clearly separate sqruff and sqlfluff test cases and allow automatically updating sqlfluff tests (create separate PR)
    * For example, sqlfluff tests should be in `sqlfluff` folder, `sqruff` specfici tests should be in `sqruff` folder
1. Use `update_test_cases.py` to copy tests from sqlfluff, add `ignored` field to all failing tests (see failed tests in `crates/lib/test.txt`)
1. Copy missing yml files from sqlfluff - `AL09.yml` and `CP02_LT01.yml`
1. Update readme(?) to add info about using `update_test_cases.py` to copy tests from sqlfluff, the script probably should be executed regularly (once per month?), this could be automated later

@benfdking what do you think about this approach for maintaining the tests?

If this is something that would be helpful I could start working on 1\., moving sqlfluff and sqruff tests into separate folders


Fank:
@benfdking I can offer you to run claude on those tests, analyse them as good as he can and then create an issue per failing rule (maybe also per dialect), so its easier to track maybe.
I really like this idea of this PR to be honest, could helped me in the beginning to understand how things work.

@DamianMaslanka5 would it be possible to add header (first line in file) comment in each file, where this file was originaly copied from?

benfdking:
I implemented this https://github.com/quarylabs/sqruff/pull/1755 which is very useful. Just tried it with trino and it's great for bringing along changes, seeing the commit in sqlfluff and implementing the equivalent parser change in sqruff.

Fank:
@benfdking would make sence, because I already fixed tsql issues in sqruff which are still broken in sqlfluff.

DamianMaslanka5:
@benfdking If by `second set of fixtures` you mean just adding a new folder for sqruff specific test, then I think that would be the best solution

benfdking:
Seeing if we can automate the writing of them in the future https://github.com/quarylabs/sqruff/pull/1779

PR #1740, opened 2025-07-03
refactor: use lazy instantiation for NodeMatcher

gvozdvmozgu:


crates/lib-dialects/src/tsql.rs
```diff
@@ -816,7 +812,7 @@ pub fn raw_dialect() -> Dialect {
     dialect.add([(
         "StringBinaryOperatorGrammar".into(),
         one_of(vec_of_erased![
-            Ref::new("ConcatSegment"), // Standard || operator
+            Ref::new("ConcatSegment"), // Standard |_| operator
```
nikolaevx:
```suggestion
            Ref::new("ConcatSegment"), // Standard || operator
```

benfdking:
This speed up is nice! @gvozdvmozgu my only question is whether we ever see a future where dialects can be fully computed at compile time. Could this whole thing not be done lazily, but upfront?

PR #1741, opened 2025-07-04
docs: document sqruff for linting sqlx queries

gvozdvmozgu:


benfdking:
Happy to add this but can you add it above the `Community` section, this can be a place where we highlight outside projects that use it?

PR #1752, opened 2025-07-05
Save sql and settings in url

DamianMaslanka5:
Having a state in url is very useful for sharing settings and sql

I think it could be also helpful to include urls with sql/settings in issues

https://github.com/user-attachments/assets/4afd2b6f-b3a9-4d04-9f92-4c0ced1cdc2d


benfdking:
nice!

PR #1770, opened 2025-07-08
feat: ORDER BY WITH FILL in clickhouse

DamianMaslanka5:
see https://github.com/sqlfluff/sqlfluff/pull/6018

I copied `OrderByClauseSegment` from ansi instead of adding `WithFillSegment` directly to order by in ansi.


benfdking:
Nice work! And thanks for the contribution! I definitely think copying is the right thing.

PR #1778, opened 2025-07-09
chore: Add ability to run dialect tests for a single dialect

DamianMaslanka5:
Running all tests takes 20s, running only tests for clickhouse takes 1.5s

Now it is possible to run `cargo test -- clickhouse`

### DIFF
--- a/crates/lib-dialects/tests/dialects.rs
+++ b/crates/lib-dialects/tests/dialects.rs
@@ -13,44 +13,17 @@ use sqruff_lib_core::parser::segments::Tables;
 use sqruff_lib_dialects::kind_to_dialect;
 use strum::IntoEnumIterator;
 
-#[derive(Default)]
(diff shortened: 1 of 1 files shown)


crates/lib-dialects/tests/dialects.rs
```diff
@@ -13,44 +13,17 @@ use sqruff_lib_core::parser::segments::Tables;
...
-    fn parse_args(&mut self, iter: impl Iterator<Item = String>) {
-        for arg in iter {
-            if arg == "--" {
-                continue;
-            }
-
-            match arg.as_str() {
-                "--list" => self.list = true,
-                "--ignored" => self.ignored = true,
-                "--no-capture" => self.no_capture = true,
```
DamianMaslanka5, line 33: **note:** I don't see these args used anywhere, so I removed them

PR #1779, opened 2025-07-09
chore: hacking script to use claude to catch up

benfdking:


Fank:
@benfdking since this merge the Python checks fail because of the claude sdk not supporting 3.9, the fail cancels all python version checks.
Noticed in #1792

PR #1785, opened 2025-07-12
feat: format and into outfile in clickhouse

DamianMaslanka5:
see https://github.com/sqlfluff/sqlfluff/pull/6065

This is not 1:1 translation of changes from sqlfluff

IMO formats should not be treated as keyword, because:
* capitalisation rules should not be applied to formats
* I want to avoid maintaining a list of formats

### DIFF
--- a/crates/lib-dialects/src/clickhouse.rs
+++ b/crates/lib-dialects/src/clickhouse.rs
@@ -12,7 +12,7 @@ use sqruff_lib_core::parser::grammar::sequence::{Bracketed, Sequence};
 use sqruff_lib_core::parser::lexer::Matcher;
 use sqruff_lib_core::parser::matchable::MatchableTrait;
 use sqruff_lib_core::parser::node_matcher::NodeMatcher;
-use sqruff_lib_core::parser::parsers::{StringParser, TypedParser};
+use sqruff_lib_core::parser::parsers::{RegexParser, StringParser, TypedParser};
 use sqruff_lib_core::parser::segments::meta::MetaSegment;
 use sqruff_lib_core::parser::types::ParseMode;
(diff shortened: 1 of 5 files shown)


crates/lib-dialects/src/clickhouse.rs
```diff
@@ -159,6 +206,60 @@ pub fn dialect() -> Dialect {
...
+        .config(|this| this.optional())
+        .to_matchable()
+        .into(),
+    )]);
+
+    clickhouse_dialect.add([
+        (
+            // https://clickhouse.com/docs/interfaces/formats
+            "FormatValueSegment".into(),
+            RegexParser::new("[a-zA-Z]*", SyntaxKind::Word)
```
DamianMaslanka5, line 236: **note:** this part is different

**question:** Is `SyntaxKind::Word` the right `SyntaxKind`?  
I am not sure which `SyntaxKind` I should use here. Should I use a different `SyntaxKind`, or add new one?

code from sqlflfuff: `OneOf(*[Ref.keyword(allowed_format) for allowed_format in FORMAT_KEYWORDS]),`

benfdking: `SyntaxKind::Word` is fine. I think the `sqlfluff` version is slightly better as it is more constraning but we could add that in the future if we needed to.

DamianMaslanka5: The problem with more constraining validation in sqlfluff is that no one updates it, the list of formats looks outdated

`select count() from system.formats` in ch returns 105 rows

sqlfluff has 81 values in `FORMAT_KEYWORDS`

PR #1924, opened 2025-08-29
fix: ignore directories

benfdking:


benfdking:
@codex can you check the issue is fixed by compiling the code and running it in a sample

PR #1951, opened 2025-09-10
fix: typo in cargo binstall windows override

MatthijsKok:


benfdking:
Nice find. Sorry you hit that!

PR #1968, opened 2025-09-14
fix(clickhouse): grammar refers to 'SETTING' which was not found in the dialect.

DamianMaslanka5:
Also added missing test for MODIFY SETTING

The same error exists in sqlfluff

I ported changes related to modify setting from sqlfluff in https://github.com/quarylabs/sqruff/pull/1843


crates/lib-dialects/test/fixtures/dialects/clickhouse/sqruff/alter_table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE x MODIFY COLUMN col MODIFY SETTING min_compress_block_size = 8192;
```
DamianMaslanka5, line 1: **note**: I initially added this file as `clickhouse/sqlfluff/alter_table_sqruff.sql`, but I just noticed that there is `sqlfluff` folder, so I added also `sqruff` folder and updated `dialects.rs` to run snapshot tests for `sqruff` folder

benfdking: Thanks for doing that! That was exactly what this is for!

PR #2069, opened 2025-12-12
fix(reflow): use relevant untaken count for indent calculations

danielchawkins:
Refactor indentation calculation logic to properly account for indent troughs when determining relevant untaken indents. Previously, the code used `untaken_indents.len()` directly, which didn't filter out irrelevant indents when a trough was present. I was getting thread errors and incorrect indentation, especially on larger .sql files, if the spacing_before = align was used, as in the sample configuration

[sqruff:layout:type:alias_expression]
spacing_before = align
align_within = select_clause
align_scope = bracketed

[oracle_parsed.sql](https://github.com/user-attachments/files/24118360/oracle_parsed.sql)

Changes:
- Add `relevant_untaken_count()` method to `IndentPoint` that mirrors the filtering logic in `IndentLine::desired_indent_units()`
- Update `lint_line_untaken_positive_indents()` to use the new method
- Update `lint_line_untaken_negative_indents()` to use the new method
- Add `.max(0)` guards to prevent negative indent values

This fixes incorrect indentation calculations in cases where indent troughs are present, ensuring that only relevant untaken indents are considered when computing desired indentation levels.

### DIFF
--- a/crates/cli-python/tests/dbt/output.stderr
+++ b/crates/cli-python/tests/dbt/output.stderr
@@ -1,6 +1,6 @@
 == [models/customers.sql] FAIL
 L:  22 | P:  16 | LT02 | Expected indent of 4 spaces. [layout.indent]
-L:  42 | P:  27 | LT02 | Expected line break and indent of 4 spaces before "on".
+L:  42 | P:  27 | LT02 | Expected line break and indent of 8 spaces before "on".
                        | [layout.indent]
 L:  42 | P:  31 | LT02 | Expected indent of 12 spaces. [layout.indent]
 L:  65 | P:  41 | LT01 | Expected only single space before "customers". Found " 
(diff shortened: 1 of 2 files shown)


benfdking:
Thanks for the contribution. I investigated this against current `main` using the attached Oracle reproduction and the alignment configuration from the PR.

This no longer appears to be needed or valid as proposed:

- The reported capacity-overflow/panic path was fixed separately by #2290, which added the Rust-specific `.max(0)` guards and a regression test.
- On current `main`, the attached file fixes without panicking, and linting the result reports no LT02 violations.
- Applying the remaining changes from this PR produces byte-for-byte identical fixed SQL for the supplied reproduction.
- Current SQLFluff retains the same behavior as sqruff: its main desired-indent calculation filters trough-related untaken indents, while the positive- and negative-indent helper paths use the full untaken-indent count. The clamp in sqruff is needed because Python safely treats negative string repetition as an empty string, whereas Rust must avoid casting a negative value to `usize`.

The proposed helper also does not exactly mirror `desired_indent_units()`: it calculates from the point's incoming balance rather than the line's effective starting/closing balance.

Given that the crash has been addressed, the supplied case no longer demonstrates incorrect final formatting, and the remaining change would diverge from SQLFluff without a focused regression case, I'm going to close this PR. If a smaller example is found where current `main` produces incorrect final SQL specifically because of trough filtering, that would be worth revisiting in a new issue or PR with a targeted test.

PR #2079, opened 2025-12-26
refactor(parser): simplify indentation config

gvozdvmozgu:


gvozdvmozgu:
It’s simpler in the sense that we no longer need a hash map. In the future, the typed config will most likely store an IndentationConfig and just pass the IndentationConfig into ParserContext, rather than creating it itself.

benfdking (approved):

PR #2080, opened 2025-12-27
feat(dialects): update dialect definitions

nikolaevx:


benfdking:
I've looked at ansi, all the fixtures look good. Is there any reason for the move away from `vec_erased`?

PR #2084, opened 2025-12-27
Fix STRING_AGG type casting style warning

benfdking:
When a shorthand cast (e.g., `id::text`) is used inside a non-CAST/CONVERT function like STRING_AGG, the rule was incorrectly flagging it as inconsistent casting style.

The issue was that when visiting a function like STRING_AGG, the rule returned `TypeCastingStyle::None` and stored it as the "prior style". Then when visiting the cast expression inside (e.g., `id::text`), it compared `None != Shorthand` and reported an inconsistency.

The fix skips storing and comparing `TypeCastingStyle::None` since it means "not a cast operation" and shouldn't be compared with actual casting styles.

Fixes issue where `SELECT STRING_AGG(id::text, ' ') FROM ou;` would incorrectly trigger CV11 with postgres dialect.

### DIFF
--- a/crates/lib/src/rules/convention/cv11.rs
+++ b/crates/lib/src/rules/convention/cv11.rs
@@ -157,6 +157,12 @@ FROM foo;
         let functional_context = FunctionalContext::new(context);
         match self.preferred_type_casting_style {
             TypeCastingStyle::Consistent => {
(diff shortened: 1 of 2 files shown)


benfdking:
I wonder if this could be more easily fixed by just looking at cast operations and not at functions as well.

PR #2317, opened 2026-02-13
fix: LT05 comment move merges comment with next line's code

garymm:
fix_long_line_with_comment used create_before with a deep leaf anchor (e.g. FunctionNameIdentifier), inserting the comment inside the wrong tree node. LT01 then removed the newline between the comment and code, producing broken SQL. Use create_after on the indent point's last segment instead, which inserts at the correct tree level.

### DIFF
--- a/crates/lib/src/rules/layout/lt05.rs
+++ b/crates/lib/src/rules/layout/lt05.rs
@@ -182,3 +182,46 @@ FROM my_table
         RootOnlyCrawler.into()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::core::config::FluffConfig;
+    use crate::core::linter::core::Linter;
+
+    /// Verifies that moving a trailing comment before its code line doesn't
+    /// merge it with the code, which would produce broken SQL since
+    /// everything after `--` becomes part of the comment.
+    #[test]
+    fn test_comment_not_merged_with_next_line() {
+        let sql = "\
+SELECT
+    COALESCE(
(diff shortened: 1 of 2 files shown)


crates/lib/src/rules/layout/lt05.rs
```diff
@@ -182,3 +182,46 @@ FROM my_table
...
+    /// everything after `--` becomes part of the comment.
+    #[test]
+    fn test_comment_not_merged_with_next_line() {
+        let sql = "\
+SELECT
+    COALESCE(
+        REGEXP_EXTRACT(project_id, '^foo-bar-(.+)$'),                -- foo-bar-baz -> baz
+        REGEXP_EXTRACT(project_id, '^qux-(.+)$')                     -- qux-corge -> corge
+    ) AS result
+FROM t
```
benfdking, line 202: Can we add this test to the LT05 fixtures with a fail and fix string rather than it being here?

garymm: The LT05 YAML fixture tests run only LT05 in isolation. This bug is caused
by a multi-rule interaction: LT05 moves the comment correctly, but then LT01 (layout.spacing) 
removes the newline between the comment and the code on a subsequent fix pass.

benfdking: I have created https://github.com/quarylabs/sqruff/issues/2352 as I think we should be able to have fixtures that test this interaction, but I have marked that as debt we can tackle later. 

Sorry about that and thanks again for the contribution.

PR #2321, opened 2026-02-14
fix(cli): prevent panic when --config contains _path/_dir keys with relative paths

### LINKED ISSUE #2320: fix(cli): --config flag panics when config contains _path/_dir keys with relative paths
## Bug Description

When using `sqruff lint --config <file>`, if the config file contains keys ending with `_path` or `_dir` (e.g., `library_path`) with relative path values, the program panics at `config.rs:444` with `config_path.unwrap()`.

## Root Cause

In `crates/cli-lib/src/lib.rs:59`, `FluffConfig::from_source(&read_file, None)` is called with `None` for the path parameter. When `ConfigLoader::get_config_elems_from_file` processes keys ending with `_path` or `_dir`, it attempts to resolve relative paths using `config_path.unwrap()`, which panics because `config_path` is `None`.

This happens because `from_source` passes both the source string and the path to `get_config_elems_from_file`, but that function explicitly forbids `(Some(path), Some(text))` — it expects either a path (to read the file) or a text string (already-read content), not both.

## Steps to Reproduce

1. Create a config file with a `_path` or `_dir` key containing a relative path:

```ini
[sqruff]
library_path = ./my_lib
```

2. Run:

```bash
sqruff lint --config my_config.cfg example.sql
```

3. The program panics.

## Expected Behavior

The relative path should be resolved relative to the config file's directory, as it works when the config is loaded via `.sqruff` file discovery.

## Proposed Fix

Add `FluffConfig::from_file(path)` that uses the existing `load_config_file` code path (which correctly passes the file path for both reading and relative path resolution), and use it in the CLI instead of manually reading the file and calling `from_source`.

See #2321.

ota2000:
## Summary

- Fix panic when using `--config` flag with a config file containing `_path` or `_dir` keys (e.g., `project_dir`, `profiles_dir`) with relative path values
- Add `FluffConfig::from_file(path)` which reads the file and passes the path for resolving relative paths via the existing `load_config_file` code path
- Use `from_file` in the CLI instead of `from_source` with a manually read string, avoiding the need to change `get_config_elems_from_file` semantics
- Add test case with a config file containing a `_path` key

## Motivation

A common pattern in dbt projects is to have a separate config file for CI (e.g., `.sqruff-ci`) with a different `target` value, while sharing the same settings:

```ini
[sqruff:templater:dbt]
project_dir = ./
profiles_dir = ./
target = dev
```

Running `sqruff lint --config .sqruff-ci` panics because `from_source` is called with `None` for the path, and `get_config_elems_from_file` cannot resolve the relative `./` paths in `project_dir` and `profiles_dir`.

## Test plan

- [x] `cargo test --package sqruff --test configure_rule` passes (includes new `config_with_path_key` test)
- [x] `cargo test --package sqruff-lib -- core::config` passes
- [x] Existing CLI tests (`config_not_found`, `dialect_override`, `ui`, `ui_json`) still pass

Closes #2320


ota2000:
Reopening — confirmed this is a real issue in production. dbt templater config commonly uses `project_dir = ./` and `profiles_dir = ./`, and when the config file is passed via `--config` (e.g., a separate CI config), sqruff panics.

crates/lib/src/core/config.rs
```diff
@@ -64,6 +64,14 @@ impl FluffConfig {
         self.reflow = ReflowConfig::from_fluff_config(self);
     }
 
+    /// from_file creates a config object from a file path. The path is used both
+    /// to read the file content and to resolve relative `_path`/`_dir` values.
+    pub fn from_file(path: &Path) -> FluffConfig {
+        let mut configs = AHashMap::new();
```
benfdking: We've moved to Hashbrown, so this will need to be updated.

crates/cli/tests/configure_rule/config_with_path_key.cfg
```diff
@@ -0,0 +1,4 @@
+[sqruff]
+dialect = ansi
+rules = core
+library_path = ./my_lib
```
benfdking: We're trying to make the config fully typed, so having this hack to test the _path is a slight mismatch. The test is great, but could we use a dbt project, potentially with a nested config, to prove that it works?

crates/lib/src/core/config.rs
ota2000: Updated — rebased on main and replaced `AHashMap` with `hashbrown::HashMap`.

crates/cli/tests/configure_rule/config_with_path_key.cfg
ota2000: Good point! Replaced the fake `library_path` key with a nested `[sqruff:templater:dbt]` section using `project_dir` and `profiles_dir`, which are actual `_dir` keys used in dbt configs.

ota2000:
Thanks for the review! Addressed both comments:

- Rebased on main to pick up the hashbrown migration, and updated `from_file` to use `HashMap` instead of `AHashMap`
- Replaced the fake `library_path` test key with a realistic nested `[sqruff:templater:dbt]` section using `project_dir` and `profiles_dir`

All tests passing (`configure_rule`, `core::config`, `ui`, `ui_json`).

PR #2322, opened 2026-02-14
fix: skip templated references in RF03 to avoid false positives

### LINKED ISSUE #2019: RF02 incorrectly raised for pyformat placeholder templates
### What Happened
sqruff raises `references.consistent` for `%(key)s` when it shouldn't, as `key` here is a template variable not a column name. sqlfluff configured the same way does not raise this.

```
15:48:45 > sqruff lint test.sql
== [test.sql] FAIL
L:   3 | P:   5 | RF03 | Unqualified reference 'key' found in single table
                       | select. [references.consistent]
L:   3 | P:   5 | RF03 | Unqualified reference 'key' found in single table
                       | select which is inconsistent with previous references.
                       | [references.consistent]
The linter processed 1 file(s).
All Finished 📜 🎉
```

### Expected Behaviour

No warnings/errors. 

```
15:48:41 > sqlfluff lint test.sql
All Finished 📜 🎉!                                                                                                                                   
```

### How to reproduce

test.sql:
```sql
SELECT
    usr.name,
    %(key)s AS key_val
FROM auth_user AS usr;

```

### Configuration
```ini
[sqruff]
dialect = postgres
rules = all
templater = placeholder

param_style = pyformat
```

ota2000:
## Summary

- Add `is_templated()` check in RF03's `validate_one_reference()` to skip template-originated references (e.g. `%(key)s` via placeholder templater)
- Add test case for `pyformat` placeholder style confirming RF03 no longer flags templated references

## Context

Closes #2019

When using `templater = placeholder` with `param_style = pyformat`, the placeholder templater replaces `%(key)s` with `key` and marks the segment as `slice_type: "templated"`. The parser then recognizes `key` as an `ObjectReference`, but RF03's `validate_one_reference()` was not checking `is_templated()`, causing false positive "unqualified reference" warnings.

This follows the same pattern used in CP01 (`cp01.rs:197`) to skip templated segments.

## Test plan

- [x] Verified the test fails without the fix (RF03 reports `Unqualified reference 'key' found in single table select`)
- [x] Verified the test passes with the fix
- [x] All existing RF03 tests continue to pass

### DIFF
--- a/crates/lib/src/rules/references/rf03.rs
+++ b/crates/lib/src/rules/references/rf03.rs
@@ -199,6 +199,10 @@ fn validate_one_reference(
         return None;
     }
 
+    if ref_.0.is_templated() {
+        return None;
+    }
+
     if standalone_aliases.contains(ref_.0.raw()) {
         return None;
     }
--- a/crates/lib/test/fixtures/rules/std_rule_cases/RF03.yml
+++ b/crates/lib/test/fixtures/rules/std_rule_cases/RF03.yml
@@ -328,3 +328,17 @@ fail_but_dont_fix_templated_table_name_qualified:
     rules:
       references.consistent:
         single_table_references: qualified
+
+test_pass_pyformat_placeholder:
+  pass_str: |
+    SELECT
+        usr.name,
+        %(key)s AS key_val
+    FROM auth_user AS usr
+  configs:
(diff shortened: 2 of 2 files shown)


benfdking:
Thanks for the contribution!

PR #2324, opened 2026-02-14
feat(rules): add LT14 rule for keyword clause newline enforcement

### LINKED ISSUE #2323: feat: add LT14 rule (keyword clause newline enforcement)
### What would you like to see?

Add a new lint rule **LT14** (`layout.keyword_newline`) that enforces each SQL keyword clause (SELECT, FROM, WHERE, JOIN, GROUP BY, HAVING, ORDER BY, LIMIT) starts on its own line.

### Motivation

SQLFluff has [LT14](https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.rules.sphinx.Rule_LT14) but sqruff does not implement it yet. This rule helps keep SQL statements readable by ensuring clauses don't share a line when other clauses already have their own lines.

### Expected behavior

**Anti-pattern:**

```sql
SELECT a
FROM foo WHERE a = 1
```

**Best practice:**

```sql
SELECT a
FROM foo
WHERE a = 1
```

The rule should respect the reflow `line_position` config:
- `alone` (default): only flags inconsistent newlines (some clauses on new lines, some not). All-on-one-line is allowed.
- `alone:strict`: always enforces newlines between clauses, even for single-line queries.

### Implementation notes

This can be implemented using the existing reflow `rebreak` system (`ReflowSequence::from_root().rebreak()`), consistent with how LT01/LT02 work. During investigation, two latent bugs were found in `identify_rebreak_spans` and `RebreakIndices::from_elements` that need to be fixed for clause-level rebreak spans to work correctly.

ota2000:
## Summary
- Add new **LT14** rule (`layout.keyword_newline`) that enforces each SQL keyword clause starts on its own line
- Implemented using the reflow `rebreak` system (`ReflowSequence::from_root().rebreak()`), consistent with LT01/LT02
- Fix two latent bugs in the reflow rebreak system:
  - `identify_rebreak_spans`: end_idx calculation for clause-level spans stopped at the first `End/Solo` block instead of scanning to the actual last block within the scope
  - `RebreakIndices::from_elements`: missing bounds checking for neighbor access in newline/pre_code scanning loops

### Rule behavior
The rule targets SELECT, FROM, WHERE, JOIN, GROUP BY, HAVING, ORDER BY, and LIMIT clauses. It respects the reflow `line_position` config:

- **`alone`** (default): flags inconsistent newlines where some clauses have newlines but others don't. All-on-one-line queries are allowed.
- **`alone:strict`**: always enforces newlines between clauses, even for single-line queries.

### Rebreak bug fixes
`identify_rebreak_spans` had a bug where clause-level `line_position_configs` spans had incorrect `end_idx`. Within a clause like `WHERE a = 1`, all blocks in the last code child (`a`, `=`, `1`) share `StackPositionType::End` at the parent level. The old code broke on the first `End` block (`a`) instead of continuing to scan to the actual last one (`1`), producing truncated spans that caused incorrect rebreak behavior.

`RebreakIndices::from_elements` had missing bounds checks where `newline_point_idx + dir` and `pre_code_point_idx + dir` could exceed array bounds when spans reached the edge of the element buffer.

## Test plan
- [x] `cargo test -p sqruff-lib --test rules -- -- "LT14.yml"` — 12 test cases pass
- [x] `cargo test -p sqruff-lib --test rules` — all rule tests pass (CV09 jinja templater failure is pre-existing)
- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy -p sqruff-lib` passes with no warnings

Closes #2323

### DIFF
--- a/crates/lib/src/templaters/placeholder.rs
+++ b/crates/lib/src/templaters/placeholder.rs
@@ -808,6 +808,6 @@ param_style = percent
         let mut linter = Linter::new(config, None, None, false);
         let result = linter.lint_string_wrapped(sql, true).unwrap().fix_string();
 
-        assert_eq!(result, "SELECT\n    a,\n    b\nFROM users WHERE a = %s\n");
+        assert_eq!(result, "SELECT\n    a,\n    b\nFROM users\nWHERE a = %s\n");
     }
 }


crates/lib/src/rules/layout/lt14.rs
```diff
@@ -0,0 +1,85 @@
...
+const CLAUSE_TYPES: SyntaxSet = SyntaxSet::new(&[
+    SyntaxKind::SelectClause,
+    SyntaxKind::FromClause,
+    SyntaxKind::WhereClause,
+    SyntaxKind::JoinClause,
+    SyntaxKind::GroupbyClause,
+    SyntaxKind::OrderbyClause,
+    SyntaxKind::HavingClause,
+    SyntaxKind::LimitClause,
+]);
```
benfdking, line 19: We should include this syntax set in the generated docs.

Issue #914, opened 2024-10-31
Error "Unknown templater: python"


quassy:
Using this `.sqruff` config:

```
[sqruff]
dialect = bigquery
templater = python
```

gives me this error:

```
thread 'main' panicked at crates/lib/src/core/linter/core.rs:58:26:
Unknown templater: python
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

even though the [default config mentions Python as one of the options](https://github.com/quarylabs/sqruff/blob/main/crates/lib/src/core/default_config.cfg#L9).

quassy:
Thanks for the quick clarification in the docs!

Can you tell me what I would need to set as param_regex to be able to use and keep Python style placeholders like `{column}`? My guess would be:

```
[sqruff]
templater = placeholder

[sqruff:templater:placeholder]
param_regex = {(?P<param_name>[\w_]+)}
```

But I'm not sure about which exact Regex syntax is used, if the specified [named capturing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Named_capturing_group) is correct and if escaping is necessary.

PR #2349, opened 2026-02-16
fix: return Result from get_templater() instead of panicking

ota2000:
## Summary
- Change `Linter::get_templater()` to return `Result<&'static dyn Templater, String>` instead of panicking on unknown templater names
- CLI validates templater early and exits with a descriptive error message listing available templaters
- LSP and WASM gracefully fall back to the raw templater when the configured templater is unavailable
- Test runners skip test cases that require unsupported templaters (e.g. `jinja`, `python` without the `python` feature)

Relates to #914

## Test plan
- [x] `cargo test -p sqruff-lib` — all tests pass, jinja/python cases are skipped
- [x] `cargo test -p sqruff --test configure_rule` — passes
- [x] `cargo fmt --all` — no formatting issues

### DIFF
--- a/crates/cli-lib/src/lib.rs
+++ b/crates/cli-lib/src/lib.rs
@@ -76,11 +76,6 @@ where
         }
     }
 
-    if let Err(e) = Linter::get_templater(&config) {
-        eprintln!("{}", e);
-        std::process::exit(1);
-    }
-
     let current_path = std::env::current_dir().unwrap();
     let ignore_file = ignore::IgnoreFile::new_from_root(&current_path).unwrap();
     let ignore_file = Arc::new(ignore_file);
@@ -131,7 +126,11 @@ where
     }
 }
 
-pub(crate) fn linter(config: FluffConfig, format: Format, collect_parse_errors: bool) -> Linter {
+pub(crate) fn linter(
+    config: FluffConfig,
+    format: Format,
+    collect_parse_errors: bool,
+) -> Result<Linter, String> {
     let formatter: Arc<dyn Formatter> = match format {
         Format::Human => {
             let output_stream = std::io::stderr().into();
--- a/crates/lib/src/core/linter/core.rs
+++ b/crates/lib/src/core/linter/core.rs
@@ -46,19 +46,18 @@ impl Linter {
         formatter: Option<Arc<dyn Formatter>>,
         templater: Option<&'static dyn Templater>,
         include_parse_errors: bool,
-    ) -> Linter {
+    ) -> Result<Linter, String> {
         let templater: &'static dyn Templater = match templater {
             Some(templater) => templater,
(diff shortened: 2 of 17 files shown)


crates/lib/src/core/linter/core.rs
```diff
@@ -49,7 +49,8 @@ impl Linter {
     ) -> Linter {
         let templater: &'static dyn Templater = match templater {
             Some(templater) => templater,
-            None => Linter::get_templater(&config),
+            None => Linter::get_templater(&config)
+                .expect("Templater should be validated before Linter::new()"),
```
benfdking: If this can error, I think we should propagate the error through this as well.
`Templater should be validated before Linter::new()` should just propagate the error upwards.

crates/lib/tests/rules.rs
```diff
@@ -150,7 +150,16 @@ fn main() {
...
+                let templater = match Linter::get_templater(linter.config()) {
+                    Ok(t) => t,
+                    Err(e) => {
+                        println!("Skipping case '{}': {}", case.name, e);
+                        *linter.config_mut() = FluffConfig::default();
+                        linter.config_mut().raw.extend(core.clone());
+                        linter.config_mut().reload_reflow();
+                        continue;
+                    }
+                };
```
benfdking, line 170: This feels unsafe and a recipe for breaking tests, especially in the way we run them . Can we make this very explicit, in that it needs an env variable to allow skipping?

crates/cli-lib/src/lib.rs
```diff
@@ -76,6 +76,11 @@ where
         }
     }
 
+    if let Err(e) = Linter::get_templater(&config) {
+        eprintln!("{}", e);
+        std::process::exit(1);
+    }
+
```
benfdking: nit: this might get more complicated with upcoming changes, so I am more keen on just failing on linter creation rather than this approach of trying to catch early.

crates/lib/src/core/linter/core.rs
ota2000: Thanks for the feedback\! `Linter::new()` now returns `Result<Linter, String>` and propagates the error via `?` instead of panicking. All call sites have been updated to handle the `Result` accordingly.

crates/lib/tests/rules.rs
ota2000: Good point. The test now panics by default when an unsupported templater is encountered. Skipping is only allowed when `SQRUFF_SKIP_UNSUPPORTED_TEMPLATERS=1` is explicitly set.

crates/cli-lib/src/lib.rs
ota2000: Agreed. Removed the early validation here. The error now propagates from `Linter::new()` which returns `Result`, so CLI commands handle it at their respective call sites.

PR #2357, opened 2026-02-22
WIP: Add support for all ClickHouse builtin functions, QUALIFY and WITH FILL FROM/TO/STEP

nastynaz:
I'm a big user of sqlfluff and got tired with the lack of ClickHouse support so I decided to implement the missing parts that are most valuable to me.

This PR introduces:
#### Support for CilckHouse QUALIFY
Filters window queries and runs after `HAVING`

#### Support for ClickHouse WITH FILL FROM / WITH FILL TO / STEP
Used for filling in timeseries data.

#### Support for all builtin ClickHouse functions (WIP. Progress: 500/1618 added)
 These are all the functions from the system.functions table along with their aliases and variants.
 ```sql
 select * from system.functions -- 1618 in total
```
Some of these are case sensitive whilst others are case in-sensitive. For the case sensitive variants I've made sure that capitalisation policies won't automatically 'fix' them to incorrect values. Case insensitive functions still adhere to the capitalisation policy.

### Summary

I'm making this a WIP PR whilst I finish the rest of the functions so that I can get feedback and make sure everything adheres to the coding standards here. Once this PR is merged `sqruff` will support much more of the ClickHouse dialect that `sqlfluff`.

It takes me a long time to check all of the aliases and variants to make sure I haven't missed any, so this might take a couple of weeks until it's done. Any feedback during this time would be much appreciated.

I also have helper Python scripts that can generate support for functions that are added in the future. I've left them out of this PR but if you think it's helpful I can add them.


benfdking:
I am going to close this as it has been a while but I would think this would need to be broken up.

PR #2529, opened 2026-04-09
fix(bigquery): support @@ system variables in lexer

bermi:
BigQuery supports both @param and @@system_variable syntax. The lexer regex only matched single @ prefixes, causing @@var to be split into two tokens. This produced unwanted line breaks during formatting (e.g. `SET foo = \n@@bar;`) and parse errors in CALL statements using system variables like @@error.message.

Widen the at_sign_literal regex from `@[a-zA-Z_][\w]*` to `@{1,2}[a-zA-Z_][\w.]*` to match both forms in a single token, with dot support for qualified names like @@error.message.

### DIFF
--- a/crates/lib-dialects/src/bigquery.rs
+++ b/crates/lib-dialects/src/bigquery.rs
@@ -36,7 +36,7 @@ pub fn dialect(config: Option<&Value>) -> Dialect {
             Matcher::string("question_mark", "?", SyntaxKind::QuestionMark),
             Matcher::regex(
                 "at_sign_literal",
-                r"@[a-zA-Z_][\w]*",
+                r"@{1,2}[a-zA-Z_][\w.]*",
                 SyntaxKind::AtSignLiteral,
             ),
         ],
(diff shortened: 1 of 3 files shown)


benfdking:
Hey 👋 

Thank you for the contribution! While this is correct, it would be better to distinguish between `at_sign_literal` and `double_at_sign_literal`, which have very different meanings. 

In BigQuery, `@param` is a query parameter while `@@error.message` is a system variable — these are semantically distinct and SQLFluff treats them as separate token types.

SQLFluff introduced this in [sqlfluff/sqlfluff#5307](https://github.com/sqlfluff/sqlfluff/pull/5307), where they define two separate lexer matchers:

```python
RegexLexer(
    "at_sign_literal",
    r"@[a-zA-Z_][\w]*",
    ...
),
RegexLexer(
    "double_at_sign_literal",
    r"@@[a-zA-Z_][\w\.]*",
    ...
),
```

Along with a dedicated `SystemVariableSegment` that wraps the `DoubleAtSignLiteralSegment`, giving `@@` variables their own node type in the AST.

It would be great to follow this pattern here — adding a separate `DoubleAtSignLiteral` syntax kind rather than combining both into a single `@{1,2}` regex. This keeps the token types semantically meaningful for downstream rules and analysis.

bermi:
@benfdking thanks for the feedback. Let me know if I got the `SystemVariableSegment` right or if we need to handle it differently.

PR #2571, opened 2026-04-20
fix: avoid panics on Jinja-templated SQL with config()/source() blocks

kimfrie:
## Summary
- Fixes the `Option::unwrap() on a None value` panic at `crates/lib/src/utils/reflow/elements.rs:272` that occurs when every newline in a `ReflowPoint` comes from a Jinja template block (e.g. the source combines `{{ config(...) }}` with `{{ ref() }}` / `{{ source() }}` on a referenced table).
- When no literal newline exists in the point, `indent_to` now emits no fix and returns the point unchanged, rather than crashing the linter run.
- Also hardens the inner `get_position_marker().unwrap()` used while matching the literal newline.

Fixes #2464.

## Reproduction
Using the minimal example from #2464 (dbt templater, bigquery dialect):

```sql
{{ config(materialized='table') }}

with raw_data as (
    select * from {{ source('conn', 'tbl') }}
               )
select *
from raw_data
```

Before: `thread '<unnamed>' panicked at crates/lib/src/utils/reflow/elements.rs:272:22: called Option::unwrap() on a None value`.
After: linter completes with ordinary LT-rule violations, no panic.

Verified locally against 12 real-world dbt/jinja models (redshift dialect, `apply_dbt_builtins = True`) that each reproduced the panic on sqruff 0.37.3 — all now lint cleanly.

## Test plan
- [x] `cargo test --release --features python -p sqruff-lib --lib reflow` — all 10 reflow tests pass
- [x] Manually re-linted 12 files from a downstream project; 0 panics vs. dozens previously

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/lib-core/src/parser/segments.rs
+++ b/crates/lib-core/src/parser/segments.rs
@@ -408,15 +408,28 @@ impl ErasedSegment {
 
         let pos_marker = self.get_position_marker().unwrap();
         if pos_marker.is_literal() && !has_descendant_source_fixes {
+            let templated_str = templated_file.templated_str.as_ref().unwrap();
+            let templated_slice = pos_marker.templated_slice.clone();
+            let source_slice = pos_marker.source_slice.clone();
+
+            // Skip patch if the position marker's slices are malformed.
+            let templated_len = templated_str.len();
+            let source_len = templated_file.source_str.len();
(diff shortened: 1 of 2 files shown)


kimfrie:
@benfdking would you help me with a review?

PR #2642, opened 2026-05-25
fix: LT01 alias alignment with multi-byte literals (#2605)

### LINKED ISSUE #2605: [Bug]: LT01 alias alignment uses byte length instead of char width for multi-byte (Cyrillic) literals
## Bug

LT01 alias alignment uses byte length instead of display width. Rows with multi-byte UTF-8 characters (Cyrillic, CJK, etc.) end up under-padded, and `sqruff lint` keeps flagging them after `sqruff fix`.

sqruff 0.38.0.

## Repro

`.sqruff`:
```ini
[sqruff]
dialect = mysql
rules = LT01

[sqruff:layout:type:alias_expression]
spacing_before = align
align_within = select_clause
```

`repro.sql`:
```sql
SELECT
    'cash' AS a,
    'наличные' AS b,
    'нет' AS c,
    'plain' AS d
FROM t
```

Run `sqruff fix repro.sql`.

## Actual

```sql
SELECT
    'cash'             AS a,
    'наличные' AS b,
    'нет'           AS c,
    'plain'            AS d
FROM t
```

## Expected (what SQLFluff produces)

```sql
SELECT
    'cash'     AS a,
    'наличные' AS b,
    'нет'      AS c,
    'plain'    AS d
FROM t
```

`наличные` is 8 Cyrillic letters but 16 UTF-8 bytes (2 bytes per char); `нет` is 3 letters / 6 bytes. The padding sqruff inserts matches byte length, not character width.

benfdking:
## Summary
- `infer_next_position` advanced column positions by **byte** length; for multi-byte UTF-8 literals (Cyrillic, CJK, …) this overshot the column, so the LT01 fix engine produced visibly misaligned padding (issue #2605).
- `get_line_pos_of_char_pos` also returned byte-indexed columns, so segment `working_line_pos` disagreed with `infer_next_position` after the fix — `sqruff lint` kept flagging fixed files.
- Both helpers now count Unicode characters, matching SQLFluff's `len(str)` semantics. Falls back to byte arithmetic when the requested position is outside the underlying string (templated content can pin positions beyond the source slice).
- Added a regression case to `test_markers_infer_next_position` covering a Cyrillic literal.

## Test plan
- [x] `bazelisk test //...` (all 31 tests pass)
- [x] Reproduced issue #2605 locally; after fix, the file matches the expected SQLFluff output and a subsequent `sqruff lint` is clean:

```sql
SELECT
    'cash'     AS a,
    'наличные' AS b,
    'нет'      AS c,
    'plain'    AS d
FROM t
```

Fixes #2605.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/lib-core/src/parser/markers.rs
+++ b/crates/lib-core/src/parser/markers.rs
@@ -175,9 +175,9 @@ impl PositionMarker {
         (
             line_no + (split.len() - 1),
             if split.len() == 1 {
-                line_pos + raw.len()
+                line_pos + raw.chars().count()
             } else {
-                split.last().unwrap().len() + 1
+                split.last().unwrap().chars().count() + 1
             },
         )
     }
@@ -339,6 +339,12 @@ mod tests {
                 raw: "\nfoo".to_string(),
                 start: 2..2,
                 end: (3, 4),
+            },
+            // Multi-byte UTF-8: positions count characters, not bytes.
+            Test {
+                raw: "'наличные'".to_string(),
+                start: 0..0,
+                end: (0, 10),
             },
         ];
 
--- a/crates/lib-core/src/templaters.rs
+++ b/crates/lib-core/src/templaters.rs
@@ -309,26 +309,42 @@ impl TemplatedFileInner {
 
     /// Get the line number and position of a point in the source file.
     /// Args:
-    ///  - char_pos: The character position in the relevant file.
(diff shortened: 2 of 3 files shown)


benfdking:
We should definitely add an end to end test with the original ask to this.

PR #2653, opened 2026-05-27
fix(duckdb): trailing comma bug

wyatt-herkamp:
## Summary
  
  DuckDB's [Friendly SQL][friendly-sql] allows trailing commas in many places, including inside function argument lists (e.g. `list_value(1, 2, 3,)`). sqruff's `duckdb` dialect was inheriting the ansi `FunctionContentsGrammar`
  unchanged, which doesn't set `allow_trailing` on the argument `Delimited` — so a stray trailing comma marks the entire function call as `Unparsable`. LT01 then no longer recognizes the function name and inserts a space before `(`,
  with the failure cascading into any nested calls.
  
  ## Changes

  - `crates/lib-dialects/src/duckdb.rs` — override `FunctionContentsGrammar`, mirroring the ansi grammar but setting `allow_trailing = true` on the inner argument `Delimited`. The two existing `allow_trailing` overrides for `ORDER BY`
  and `GROUP BY` already follow the same pattern.
  - `crates/lib-dialects/test/fixtures/dialects/duckdb/sqruff/function_trailing_comma.{sql,yml}` — new fixture parsing `list_value(1, 2, 3,)` and `COALESCE(a, b, c,)` cleanly (no `unparsable` nodes; trailing `comma` token preserved
  before `end_bracket`).

  ## Test plan
  
  - [x] `cargo test -p sqruff-lib-dialects` passes
  - [x] New fixture parses without `unparsable` segments
  - [x] No regressions across other dialect fixtures

  [friendly-sql]: https://duckdb.org/docs/current/sql/dialect/friendly_sql

### DIFF
--- a/crates/lib-dialects/src/duckdb.rs
+++ b/crates/lib-dialects/src/duckdb.rs
@@ -296,7 +296,7 @@ pub fn raw_dialect() -> Dialect {
     );
 
     // DuckDB allows trailing commas in function argument lists, e.g.
-    // `struct_pack(a := 1, b := 2,)`. Mirror the ansi `FunctionContentsGrammar`
+    // `list_value(1, 2, 3,)`. Mirror the ansi `FunctionContentsGrammar`
     // but enable `allow_trailing` on the inner argument `Delimited`.
     duckdb_dialect.replace_grammar(
         "FunctionContentsGrammar",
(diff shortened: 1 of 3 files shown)


crates/lib-dialects/src/duckdb.rs
```diff
@@ -295,6 +295,83 @@ pub fn raw_dialect() -> Dialect {
         .to_matchable(),
     );
 
+    // DuckDB allows trailing commas in function argument lists, e.g.
+    // `struct_pack(a := 1, b := 2,)`. Mirror the ansi `FunctionContentsGrammar`
```
benfdking: Can we just update this slightly misleading comment. I don't think struct_pack with a trailing comma works. I think it's because of the `:=` though.

benfdking (commented):
Thanks for the nice, well-scoped fix! Just one little misleading thing.

crates/lib-dialects/src/duckdb.rs
wyatt-herkamp: Fixed

benfdking: Amazing thank you very much!

PR #2672, opened 2026-06-06
fix(postgres): parse ON CONFLICT after a FROM-less INSERT ... SELECT

skyf0l:
## Summary

In the Postgres dialect, `INSERT … SELECT … ON CONFLICT …` is reported as an `Unparsable section` whenever the `SELECT` source has **no `FROM` clause**. The same statement parses fine with a `VALUES` source, or with a `SELECT` that *does* have a `FROM` clause — so the upsert grammar is correct; the bare select target list simply doesn't stop at `ON CONFLICT` and greedily consumes the `ON` token.

This is valid PostgreSQL: an `INSERT` source may be a `query` (a `SELECT`), and `ON CONFLICT` applies to all three source forms (`DEFAULT VALUES`, `VALUES`, `query`). See the [INSERT synopsis](https://www.postgresql.org/docs/current/sql-insert.html).

## Reproduction

```sql
-- dialect: postgres

-- ❌ Unparsable section (at `ON`)
INSERT INTO t (a, b) SELECT 1, 2 ON CONFLICT (a, b) DO NOTHING;
INSERT INTO t (a, b) SELECT 1, 2 ON CONFLICT (a) DO UPDATE SET b = 1;

-- ✅ already parsed fine
INSERT INTO t (a, b) VALUES (1, 2) ON CONFLICT (a, b) DO NOTHING;
INSERT INTO t (a, b) SELECT 1, c.id FROM unnest('{}'::uuid[]) c(id) ON CONFLICT (a, b) DO NOTHING;
```

A common real-world case is a writable CTE inserting unnested rows:

```sql
WITH ins AS (
    INSERT INTO link (parent_unid, child_unid)
    SELECT $1, unnest($2::uuid[])
    ON CONFLICT (parent_unid, child_unid) DO NOTHING
)
SELECT 1;
```

## Root cause

The Postgres `SelectClauseSegment` parses its target list in `ParseMode::GreedyOnceStarted` with an inline `terminators` list (`FROM`, `WHERE`, `LIMIT`, `ORDER BY`, …) that does **not** include `ON CONFLICT`. With no `FROM` clause to terminate at, the greedy target list swallows the trailing `ON` and the statement becomes unparsable. When a `FROM` clause is present the select clause terminates there, which is why the bug only affects FROM-less selects.

## Fix

Add `ON CONFLICT` as a terminator of the Postgres select target list. The two-keyword sequence is intentional so it does **not** affect `SELECT DISTINCT ON (…)` or `JOIN … ON …`.

```diff
                  Ref::keyword("LIMIT").to_matchable(),
                  Ref::keyword("OVERLAPS").to_matchable(),
                  Ref::new("SetOperatorSegment").to_matchable(),
                  Sequence::new(vec![
                      Ref::keyword("WITH").to_matchable(),
                      Ref::keyword("NO").optional().to_matchable(),
                      Ref::keyword("DATA").to_matchable(),
                  ])
                  .to_matchable(),
                  Ref::new("WithCheckOptionSegment").to_matchable(),
+                Sequence::new(vec![
+                    Ref::keyword("ON").to_matchable(),
+                    Ref::keyword("CONFLICT").to_matchable(),
+                ])
+                .to_matchable(),
              ];
              this.parse_mode(ParseMode::GreedyOnceStarted);
```

(`crates/lib-dialects/src/postgres.rs`, in the `SelectClauseSegment` `replace_grammar` terminators.)

## Tests

Added a sqruff-native fixture covering `DO NOTHING`, `DO UPDATE`, and the writable-CTE form:

- `crates/lib-dialects/test/fixtures/dialects/postgres/sqruff/insert_select_on_conflict.sql`
- `crates/lib-dialects/test/fixtures/dialects/postgres/sqruff/insert_select_on_conflict.yml` (generated via `env UPDATE_EXPECT=1 cargo test`)

The generated parse tree contains real `insert_statement` / `conflict_target` / `conflict_action` / `with_compound_statement` nodes and **no `unparsable` segments**.

`cargo test -p sqruff-lib-dialects` passes (all dialect fixtures, 0 failed), and **no existing fixtures changed** — the grammar tweak doesn't perturb any other parse tree.

Manually re-verified after the fix:

- ✅ now parse: `INSERT … SELECT … ON CONFLICT DO NOTHING`, `… DO UPDATE`, and the writable-CTE `INSERT … SELECT unnest(…) … ON CONFLICT` form
- ✅ no regression: `SELECT DISTINCT ON (a) a, b FROM t`, `SELECT a.x FROM a JOIN b ON a.id = b.id`, `INSERT … VALUES … ON CONFLICT`, `INSERT … SELECT … FROM … ON CONFLICT`, plain
`SELECT 1, 2`, `SELECT a b FROM t`

### DIFF
--- a/crates/lib-dialects/src/postgres.rs
+++ b/crates/lib-dialects/src/postgres.rs
@@ -2309,6 +2309,11 @@ pub fn raw_dialect() -> Dialect {
                 ])
                 .to_matchable(),
                 Ref::new("WithCheckOptionSegment").to_matchable(),
+                Sequence::new(vec![
+                    Ref::keyword("ON").to_matchable(),
+                    Ref::keyword("CONFLICT").to_matchable(),
+                ])
+                .to_matchable(),
             ];
             this.parse_mode(ParseMode::GreedyOnceStarted);
         })
--- a/crates/lib-dialects/src/postgres.rs
+++ b/crates/lib-dialects/src/postgres.rs
@@ -2314,6 +2314,7 @@ pub fn raw_dialect() -> Dialect {
                     Ref::keyword("CONFLICT").to_matchable(),
                 ])
                 .to_matchable(),
+                Ref::keyword("RETURNING").to_matchable(),
             ];
             this.parse_mode(ParseMode::GreedyOnceStarted);
         })
@@ -6541,6 +6542,7 @@ pub fn raw_dialect() -> Dialect {
                                 Sequence::new(vec![
                                     one_of(vec![
                                         Ref::new("ColumnReferenceSegment").to_matchable(),
+                                        Ref::new("FunctionSegment").to_matchable(),
                                         Bracketed::new(vec![
                                             Ref::new("ExpressionSegment").to_matchable(),
                                         ])
--- a/crates/lib-dialects/test/fixtures/dialects/postgres/sqlfluff/insert.sql
+++ b/crates/lib-dialects/test/fixtures/dialects/postgres/sqlfluff/insert.sql
@@ -64,3 +64,36 @@ INSERT INTO abc (foo, bar)
 SELECT foo, bar FROM baz
 RETURNING quux
 ;
+
+INSERT INTO tbl_a (
+   val1
+ , val2
+)
+SELECT val1
+     , val2
+FROM tbl_2
+ON CONFLICT (
+    val1
+  , COALESCE(val2, '')
+)
+DO NOTHING;
+
+INSERT INTO prompt_variants (
+    test,
+    test2
+)
+SELECT
+    test,
+    test2
+RETURNING
+  test,
+  test2;
+
+INSERT INTO
+  baz (state, state_changed_at, instance_id)
+SELECT
+  1, 2, 3
+ON CONFLICT (instance_id) DO UPDATE
+SET
+  state = foo,
+  state_changed_at = bar;
--- a/crates/lib-dialects/test/fixtures/dialects/postgres/sqlfluff/insert.yml
+++ b/crates/lib-dialects/test/fixtures/dialects/postgres/sqlfluff/insert.yml
@@ -791,3 +791,149 @@ file:
       - column_reference:
         - naked_identifier: quux
 - statement_terminator: ;
+- statement:
+  - insert_statement:
+    - keyword: INSERT
+    - keyword: INTO
+    - table_reference:
+      - naked_identifier: tbl_a
+    - bracketed:
+      - start_bracket: (
+      - column_reference:
+        - naked_identifier: val1
+      - comma: ','
+      - column_reference:
+        - naked_identifier: val2
+      - end_bracket: )
+    - select_statement:
+      - select_clause:
+        - keyword: SELECT
+        - select_clause_element:
+          - column_reference:
+            - naked_identifier: val1
+        - comma: ','
+        - select_clause_element:
+          - column_reference:
+            - naked_identifier: val2
+      - from_clause:
+        - keyword: FROM
+        - from_expression:
+          - from_expression_element:
+            - table_expression:
+              - table_reference:
+                - naked_identifier: tbl_2
+    - keyword: ON
+    - keyword: CONFLICT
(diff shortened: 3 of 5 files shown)


benfdking:
Thanks for the fix — the `ON CONFLICT` terminator on `SelectClauseSegment` is correct and the parse tree it produces matches SQLFluff's. A couple of suggestions to make this a faithful port of upstream:

**1. This mirrors SQLFluff [#6864](https://github.com/sqlfluff/sqlfluff/pull/6864) ("Terminate SELECT with ON CONFLICT") — worth porting it directly.** Rather than a bespoke fixture, it'd be cleaner to sync upstream's change into our already-ported `crates/lib-dialects/test/fixtures/dialects/postgres/sqlfluff/insert.sql` (which is currently stale relative to upstream) and regenerate the `.yml`. That keeps our SQLFluff-derived fixtures tracking theirs. The custom `sqruff/insert_select_on_conflict.sql` adds nice extra coverage (`DO NOTHING` + writable CTE), so keeping it as a supplement is fine — but the upstream case (`INSERT INTO baz … SELECT 1, 2, 3 ON CONFLICT (instance_id) DO UPDATE SET …`) should be ported too.

**2. There's a sibling fix you'll want to grab at the same time: SQLFluff [#6832](https://github.com/sqlfluff/sqlfluff/pull/6832) ("Postgres: Fix RETURNING Select terminator").** Upstream fixed `ON CONFLICT` and `RETURNING` as two separate PRs, and they ended up as adjacent terminators in `SelectClauseSegment`:

```python
Sequence("ON", "CONFLICT"),
"RETURNING",
```

We only have the `ON CONFLICT` half here. The exact same FROM-less greedy-consume bug still affects `RETURNING`, so this still fails to parse:

```sql
INSERT INTO prompt_variants (test, test2)
SELECT test, test2
RETURNING test, test2;
```

(SQLFluff covers this in `insert.sql` via the `prompt_variants` case from #6832.) Adding `Ref::keyword("RETURNING")` alongside the new `ON CONFLICT` terminator — and porting that fixture case — would close both gaps and bring `SelectClauseSegment` fully in line with upstream.

Note: SQLFluff's terminator list also includes `MetaCommandQueryBufferSegment`, but that's psql meta-command (`\g`) support we don't have, so skipping it is fine.

skyf0l:
Thanks! Addressed both points:

1. **`RETURNING` terminator ([sqlfluff#6832](https://github.com/sqlfluff/sqlfluff/pull/6832)) + `ON CONFLICT` sync ([sqlfluff#6864](https://github.com/sqlfluff/sqlfluff/pull/6864))** — added `RETURNING` as a `SelectClauseSegment` terminator right after the `ON CONFLICT` one, matching upstream's adjacency. Synced the stale `sqlfluff/insert.sql` up to upstream (the `prompt_variants … SELECT … RETURNING` case from [sqlfluff#6832](https://github.com/sqlfluff/sqlfluff/pull/6832) and the `baz … SELECT 1, 2, 3 ON CONFLICT (instance_id) DO UPDATE SET …` case from [sqlfluff#6864](https://github.com/sqlfluff/sqlfluff/pull/6864)) and regenerated the `.yml`. Kept the custom `insert_select_on_conflict` fixture as a supplement and gave it a `RETURNING` case too.

2. **One extra:** upstream's `insert.sql` also contains `ON CONFLICT (val1, COALESCE(val2, ''))`, a bare function-call conflict target that our `ConflictTargetSegment` couldn't parse (it only allowed `ColumnReference` or a parenthesised expression). Added `FunctionSegment`, mirroring the existing `IndexElementSegment`, so the file fully syncs — happy to drop that case if you'd rather keep this PR scoped strictly to the two terminators.

All postgres fixtures parse with no unparsable segments, and only `insert.yml` changed, so the grammar tweaks don't perturb any other parse tree. Skipped `MetaCommandQueryBufferSegment` as you noted.

skyf0l:
Hi @benfdking, could you check that one please, this is blocking the sqruff integration in some of my projects, thanks a lot :smiley:

PR #2676, opened 2026-06-06
fix(ST03): exempt data-modifying CTEs from unused_cte

skyf0l:
## What

ST03 (`structure.unused_cte`) reported a violation for any CTE whose name never appears in a `TableReference`, even when the CTE body is a data-modifying statement (`INSERT` / `UPDATE` / `DELETE` / `MERGE`). In Postgres a data-modifying CTE is executed for its side effects whether or not the primary query references it, so such a CTE is never "unused".

This is a false positive on a common, idiomatic pattern: running several deletes/inserts as one atomic statement / round-trip.

```sql
WITH deleted_a AS (
    DELETE FROM child_a WHERE parent_id = 1
),
deleted_b AS (
    DELETE FROM child_b WHERE parent_id = 1
)
DELETE FROM parent WHERE id = 1
```

Before this change all three of these CTEs were reported as unused, and there is no inline way to silence it.

## Why

SQLFluff (which sqruff tracks for rule behaviour) already exempts these via `_is_data_modifying_cte`. sqruff simply never ported that exemption. This PR brings sqruff in line.

## How

In `RuleST03::eval`, before pushing a violation for a remaining CTE, inspect its `cte_definition_segment` with `recursive_crawl` for `InsertStatement` / `UpdateStatement` / `DeleteStatement` / `MergeStatement`. If found, skip the CTE.

## Tests

Added Postgres fixtures to `ST03.yml`:

- `test_pass_postgres_data_modifying_delete_cte` — single unreferenced DELETE CTE passes.
- `test_pass_postgres_data_modifying_multiple_delete_ctes` — several unreferenced DELETE CTEs in one statement pass.
- `test_pass_postgres_data_modifying_insert_cte` — unreferenced INSERT CTE passes.
- `test_pass_postgres_data_modifying_update_cte` — unreferenced UPDATE CTE passes.
- `test_fail_postgres_unused_select_cte_alongside_data_modifying` — a genuinely unused SELECT CTE next to an exempt data-modifying CTE is still flagged, so the exemption does not over-reach.

All existing ST03 cases continue to pass.

### DIFF
--- a/crates/lib/src/rules/structure/st03.rs
+++ b/crates/lib/src/rules/structure/st03.rs
@@ -89,6 +89,32 @@ FROM cte1
         for name in remaining_ctes.values() {
             let tmp = RefCell::borrow(&query.inner);
             let cte = RefCell::borrow(&tmp.ctes[name].inner);
+
+            // Data-modifying CTEs (INSERT/UPDATE/DELETE/MERGE) are executed for
+            // their side effects in dialects like Postgres, whether or not the
+            // primary query references them, so they are never "unused".
+            // Mirrors SQLFluff ST03's `_is_data_modifying_cte` exemption.
+            if let Some(def) = &cte.cte_definition_segment {
+                let is_data_modifying = !def
+                    .recursive_crawl(
+                        const {
+                            &SyntaxSet::new(&[
+                                SyntaxKind::InsertStatement,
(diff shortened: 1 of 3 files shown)


skyf0l:
Hi @benfdking, could you check that one please, this is blocking the sqruff integration in some of my projects, thanks a lot :smiley:

PR #2742, opened 2026-06-14
feat: add semantic token highlighting to the playground and VS Code

benfdking:
## Summary

Add LSP semantic token support and use it to provide syntax-aware SQL highlighting in both the web playground and the VS Code extension.

## What changed

- Add a semantic-token classifier that maps sqruff CST `SyntaxKind` values into LSP token categories including keywords, strings, numbers, comments, operators, functions, types, variables, parameters, properties, and macros.
- Generate delta-encoded LSP semantic tokens with UTF-16 positions and split multiline tokens into valid per-line ranges.
- Advertise and handle `textDocument/semanticTokens/full` in the native and browser LSP implementations.
- Expose semantic-token generation through the WASM bindings used by the playground.
- Register a Monaco semantic-token provider in the playground and explicitly enable semantic highlighting.
- Forward semantic-token requests through the VS Code browser worker.
- Update the Bazel Rust/WASM dependency graph and generated lock metadata for the new integration.

## Behaviour and compatibility

- Unknown syntax kinds remain unhighlighted rather than receiving an incorrect classification.
- Parse failures return an empty token list so highlighting cannot break editor operation.
- Token positions follow the LSP UTF-16 coordinate requirements, including non-ASCII input.

## Tests

- Unit coverage for classification, delta encoding, multiline tokens, and UTF-16 positions.
- Protocol-level browser-worker coverage for semantic-token responses.
- Playground Playwright coverage that verifies raw semantic-token data and distinct rendered keyword/variable colours.
- VS Code Playwright coverage that verifies distinct rendered semantic keyword/variable colours in the installed extension.
- `bazelisk test //...` — 33/33 tests pass.

### DIFF
--- a/crates/lsp/src/semantic.rs
+++ b/crates/lsp/src/semantic.rs
@@ -1,0 +1,1437 @@
+//! Builds LSP semantic tokens from sqruff's concrete syntax tree.
+//!
+//! sqruff's parser already disambiguates the grammar while parsing, emitting a
+//! distinct [`SyntaxKind`] for each leaf segment (keyword, literal, operator,
+//! function name, ...). That means highlighting reduces to a flat
+//! `SyntaxKind -> Highlight` lookup ([`classify`]) over the leaf segments, with
+//! no separate query language. Kinds without syntax highlighting are explicitly
+//! listed so new dialect kinds require a deliberate classification.
+
+use lsp_types::{SemanticToken, SemanticTokenType, SemanticTokensLegend};
+use sqruff_lib::core::linter::core::Linter;
+use sqruff_lib_core::dialects::syntax::SyntaxKind;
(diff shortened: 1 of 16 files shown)


crates/lsp/src/semantic.rs
```diff
@@ -0,0 +1,410 @@
...
+            Highlight::Property
+        }
+
+        // Generic identifiers (column / table / object references resolve to
+        // these at the leaf level).
+        NakedIdentifier | QuotedIdentifier | Identifier | NakedIdentifierAll => Highlight::Variable,
+
+        // Everything else (punctuation, brackets, structural nodes, meta, and
+        // any kind we have not classified yet) is left un-highlighted.
+        _ => return None,
```
benfdking: Can we explicitely list them here rather than just match anything else so that if we add new types it needs to make a decision.

PR #2748, opened 2026-06-15
fix: handle postgres pgvector table layout

### LINKED ISSUE #2747: [Bug]: PostgreSQL CREATE TABLE with pgvector still loses indentation
### What happened?

This appears to be a follow-up to #2070. That issue is closed as completed, but the same core formatting problem still reproduces for me on `sqruff 0.38.0` with the PostgreSQL dialect.

When a `CREATE TABLE` statement contains a pgvector-style `vector(...)` column type, `sqruff fix` removes indentation from the table body and inserts spaces into PostgreSQL type syntax. In a slightly larger table, it also removes expected continuation indentation from table constraints and inserts a space before the final semicolon after `PARTITION BY LIST (...)`.

Input:

```sql
CREATE EXTENSION IF NOT EXISTS vector;

CREATE TABLE IF NOT EXISTS example_embeddings (
    id BIGSERIAL NOT NULL,
    source_id BIGINT NOT NULL,
    content TEXT NOT NULL,
    embedding vector(1536),
    labels TEXT[],
    PRIMARY KEY (source_id, id),
    CONSTRAINT fk_example_source
        FOREIGN KEY (source_id)
        REFERENCES example_sources(source_id)
        ON DELETE CASCADE
) PARTITION BY LIST (source_id);
```

Command:

```console
sqruff fix /tmp/sqruff-pgvector-repro.sql
```

Actual output:

```sql
CREATE EXTENSION IF NOT EXISTS vector;

CREATE TABLE IF NOT EXISTS example_embeddings (
id BIGSERIAL NOT NULL,
source_id BIGINT NOT NULL,
content TEXT NOT NULL,
embedding vector (1536),
labels TEXT [],
PRIMARY KEY (source_id, id),
CONSTRAINT fk_example_source
FOREIGN KEY (source_id)
REFERENCES example_sources (source_id)
ON DELETE CASCADE
) PARTITION BY LIST (source_id) ;
```

The reported violations include:

```text
L:   4 | P:   1 | LT02 | Line should not be indented. [layout.indent]
L:   7 | P:  21 | LT01 | Expected single whitespace between "vector" and "(".
L:   8 | P:  16 | LT01 | Expected single whitespace between "TEXT" and "[".
L:  14 | P:  32 | LT01 | Expected single whitespace between ")" and ";".
```

### Expected behavior

The table body should remain indented, table constraint continuation lines should remain nested, and PostgreSQL type syntax should not be rewritten with extra spaces:

```sql
CREATE EXTENSION IF NOT EXISTS vector;

CREATE TABLE IF NOT EXISTS example_embeddings (
    id BIGSERIAL NOT NULL,
    source_id BIGINT NOT NULL,
    content TEXT NOT NULL,
    embedding vector(1536),
    labels TEXT[],
    PRIMARY KEY (source_id, id),
    CONSTRAINT fk_example_source
        FOREIGN KEY (source_id)
        REFERENCES example_sources(source_id)
        ON DELETE CASCADE
) PARTITION BY LIST (source_id);
```

### Version

```console
$ sqruff --version
sqruff 0.38.0
```

### Config

```ini
[sqruff]
dialect = postgres
```

### Checked in Playground

- [ ] I have checked this bug in the playground.

mtibben:
PostgreSQL pgvector table formatting should preserve the expected layout for CREATE TABLE statements that combine extension-style vector columns, array types, table constraints, and partition clauses.

The current Postgres formatting path treats pgvector-style `vector(...)` and adjacent bracketed syntax too broadly, so reflow inserts spaces into PostgreSQL type/reference syntax and removes expected continuation indentation from table constraints. This change gives the parser and reflow config more precise segment types for these constructs so `vector(1536)`, `TEXT[]`, and `REFERENCES table(column)` keep their PostgreSQL shape while table bodies retain their indentation.

Closes #2747

### DIFF
--- a/crates/lib-core/src/dialects/syntax.rs
+++ b/crates/lib-core/src/dialects/syntax.rs
@@ -54,6 +54,7 @@ pub enum SyntaxKind {
     WithCompoundStatement,
     CommonTableExpression,
     CTEColumnList,
+    ReferencedColumnList,
     TriggerReference,
     TableConstraint,
     JoinOnCondition,
--- a/crates/lib-core/src/dialects/syntax.rs
+++ b/crates/lib-core/src/dialects/syntax.rs
@@ -134,6 +134,7 @@ pub enum SyntaxKind {
     IntervalExpression,
     ArrayType,
     SizedArrayType,
+    ArrayTypeSuffix,
     SelectStatement,
     OverlapsClause,
     SelectClause,
--- a/crates/lib-dialects/src/postgres.rs
+++ b/crates/lib-dialects/src/postgres.rs
@@ -285,14 +285,7 @@ fn build_datatype_segment_grammar(pgvector: bool) -> Matchable {
         ])
         .to_matchable(),
         one_of(vec![
-            AnyNumberOf::new(vec![
-                Bracketed::new(vec![
-                    Ref::new("ExpressionSegment").optional().to_matchable(),
-                ])
-                .config(|this| this.bracket_type("square"))
-                .to_matchable(),
-            ])
-            .to_matchable(),
+            Ref::new("ArrayTypeSuffixSegment").to_matchable(),
             Ref::new("ArrayTypeSegment").to_matchable(),
             Ref::new("SizedArrayTypeSegment").to_matchable(),
         ])
@@ -1266,6 +1259,23 @@ pub fn raw_dialect() -> Dialect {
                 Ref::new("DateTimeTypeIdentifier").optional().to_matchable(),
                 Ref::new("QuotedLiteralSegment").to_matchable(),
             ])
+            .to_matchable()
+        })
+        .to_matchable()
+        .into(),
+    )]);
+
+    postgres.add([(
+        "ArrayTypeSuffixSegment".into(),
+        NodeMatcher::new(SyntaxKind::ArrayTypeSuffix, |_| {
+            AnyNumberOf::new(vec![
+                Bracketed::new(vec![
+                    Ref::new("ExpressionSegment").optional().to_matchable(),
+                ])
+                .config(|this| this.bracket_type("square"))
+                .to_matchable(),
+            ])
+            .config(|this| this.min_times(1))
             .to_matchable()
         })
         .to_matchable()
--- a/crates/lib-dialects/src/postgres.rs
+++ b/crates/lib-dialects/src/postgres.rs
@@ -277,7 +277,11 @@ fn build_datatype_segment_grammar(pgvector: bool) -> Matchable {
             Ref::new("WellKnownTextGeometrySegment").to_matchable(),
             Ref::new("DateTimeTypeIdentifier").to_matchable(),
             Sequence::new(vec![one_of(known_types).to_matchable()]).to_matchable(),
-            Ref::new("DatatypeIdentifierSegment").to_matchable(),
+            Sequence::new(vec![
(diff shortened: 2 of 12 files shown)


mikills:
Verified locally. The pgvector/Postgres fix looks good, but I think this needs one change before merge.

  The new global config:
  `[sqruff:layout:type:start_square_bracket]`
  `spacing_before = touch:inline`

  regresses array literals outside Postgres. For example, on this PR:

  `SELECT [1, 2, 3] AS arr`

  is fixed to:

  `SELECT[1, 2, 3] AS arr`

  for BigQuery, SparkSQL, and DuckDB. On main, those preserve the space.

  Could we scope the `TEXT[]` behavior to the Postgres array datatype suffix instead of changing `start_square_bracket`
  globally? A regression fixture for array literal spacing would be useful too.

  The rest looks solid: `vector(1536)`, `TEXT[]`, and `REFERENCES table(col)` are handled much better.

mtibben:
Thanks for catching this. I agreed that the `start_square_bracket` layout change was too broad, so I removed the global spacing rule and scoped the touch-before behavior to a new Postgres `array_type_suffix` segment instead.

I also added regression coverage for `SELECT [1, 2, 3] AS arr` under BigQuery, SparkSQL, and DuckDB, and kept the pgvector/Postgres fixture covering `TEXT[]`, `vector(1536)`, and `REFERENCES table(col)`.

benfdking:
This is already partially solved with the correct config as well. Did we consider that? 

https://playground.quary.dev/?secondary=Cst#eNptUdFqwkAQ/JUlTwq+lNI+WPpwxlWOphe5XKFiJKg5gzReYu4shdJ/78ZEbYP3uDNzOzvz7dlD7g09XyJTCPiuUEQ8FMAnIEJFAx6pCD71xhXVU2xi0zIVGwXYYemv1b7MdaL3a52mO5NZ6MUG6O1SGPFphJKz4KQQb0EwaDBbHKuNThoKF4T9xzeFcdo4UGSui102tQ57dw/3j/0WzFdrnduTbrFsZzPJX5mcwwvOoXfZPCCDZ5UfikhJVhvZfiTnkxpqw6jfJJTIp6LzT/9KkDhBicLHaywNzd6kU+JjDJCC9VnkszHGpg8zJhVXdRujOQQU8d9VVIY38Kx2rg6aKlzYQ3XcbpexSXernOKAZygL67JK27q4Fh+26PCMkaDMEldlexK46khXllkT52Xg/fwCjKStsg== the right conig

Looking at your query, though, I think there are still some improvements that we should make. It might be worth just breaking them up a little.

mtibben:
> This is already partially solved with the correct config as well. Did we consider that?
> 
> Looking at your query, though, I think there are still some improvements that we should make. It might be worth just breaking them up a little.

Ah I see - with `[sqruff:dialect:postgres] pgvector = true`, the pgvector datatype parsing side is already covered, so this PR is bundling that config-dependent behavior with a few separate layout issues.

I’ve split the remaining formatter changes into smaller draft PRs:

- #2780: Postgres array datatype suffix spacing, e.g. `text[]`, while avoiding changes to array literal spacing in other dialects.
- #2781: Postgres referenced column list spacing, e.g. `REFERENCES parent(id)`.
- #2782: Named Postgres table constraint body indentation after `CONSTRAINT name`.

I think we can treat this PR as superseded by those smaller PRs and review whichever fixes still make sense independently.

PR #2781, opened 2026-06-21
fix: preserve postgres reference column list spacing

mtibben:
PostgreSQL foreign-key references should preserve the table-to-column-list form used by `REFERENCES parent(id)`.

The referenced column list currently parses as a generic bracketed list, so layout treats it like ordinary bracketed syntax and can insert a space between the referenced table and its column list. This adds a dedicated `referenced_column_list` segment for Postgres reference definitions so layout can keep that relationship tight without changing generic bracket handling.

The change applies to column and table foreign-key references through the shared `ReferenceDefinitionGrammar`.

### DIFF
--- a/crates/lib-core/src/dialects/syntax.rs
+++ b/crates/lib-core/src/dialects/syntax.rs
@@ -54,6 +54,7 @@ pub enum SyntaxKind {
     WithCompoundStatement,
     CommonTableExpression,
     CTEColumnList,
+    ReferencedColumnList,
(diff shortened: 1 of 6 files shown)


benfdking:
Thank you for raising this! This looks correct but 'REFERENCES table(column)' is part of the ansi standard. Could we move it into there so that we keep this consistent across the dialects.

benfdking:
I've put this into ansi. So closing this.

PR #2882, opened 2026-07-10
feat(sqlite): parse CTE materialization hints

karimudev:
## Summary

SQLite has support for non-standard SQL syntax `AS MATERIALIZED` and `AS NOT MATERIALIZED` in CTE definitions. (see: [The WITH Clause: Materialization Hints](https://sqlite.org/lang_with.html#materialization_hints)).
This syntax originates from Postgres. 

sqruff parses these materialization hints properly for the Postgres dialect (see: [postgres.rs:7467](https://github.com/quarylabs/sqruff/blob/v0.39.0/crates/lib-dialects/src/postgres.rs#L7467)) but not for SQLite.

## Changes

- Adds SQLite specific logic in the dialect parser for CTE definition
- Adds tests `with_cte_as_materialized` and `with_cte_as_not_materialized` for the SQLite dialect

## Validation

- `cargo test -p sqruff-lib-dialects --test dialects`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,8 +12,8 @@ You'll also need to set up a python virtualenv and install dependencies as follo
 
 ```bash
 virtualenv .venv
-source ./venv/bin/activate
-make python_install
+source .venv/bin/activate
+make python_sync
 ```
 
 ### Running Tests
--- a/crates/lib-dialects/src/sqlite.rs
+++ b/crates/lib-dialects/src/sqlite.rs
@@ -1129,6 +1129,25 @@ pub fn raw_dialect() -> Dialect {
         .to_matchable(),
     );
 
+    sqlite_dialect.replace_grammar(
+        "CTEDefinitionSegment",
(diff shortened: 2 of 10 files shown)


karimudev:
Side-note, as I'm not sure where to put this: the same issue seems to exist in sqfluff. I'm not really sure if the two projects share fixes or not, so I don't know if it is worth opening an issue/pr there as well.

nikolaevx:
This should be sent to SQLFluff as a separate issue/PR as well. These are separate projects, and fixes are not automatically shared between them.

We should probably wait until the change is accepted in SQLFluff first, and then apply the same fix here.

karimudev:
PR title had a space in front of it, probably why `PR Title Check` failed. I removed it.

Here's the PR on SQLFluff's side: [sqlfluff/sqlfluff#8126](https://github.com/sqlfluff/sqlfluff/pull/8126)

karimudev:
PR has now been merged on SQLFluff's side

PR #3046, opened 2026-07-30
fix(trino): catch up with SQLFluff and stop panicking on common statements

### LINKED ISSUE #3017: Parser panics on CREATE TABLE / INSERT in the Trino dialect (grammar refs to keywords Trino removed)
## What happens

Parsing everyday statements in the Trino dialect panics instead of producing a parse tree or a clean error.

```rust
let dialect = kind_to_dialect(&DialectKind::Trino, None).unwrap();
let lexer = Lexer::from(&dialect);
let parser = Parser::from(&dialect);
let tokens = lexer.lex(&tables, "create table".to_string());
parser.parse(&tables, &tokens.0).unwrap(); // panics
```

- `create table` panics with `Grammar refers to 'TRANSIENT' which was not found in the dialect` (grammar.rs:133)
- `insert into t values (1)` panics with `Grammar refers to 'OVERWRITE' which was not found in the dialect` (grammar.rs:145)
- `alter table t ...` reaches the same panic for `SEQUENCE`

I hit these while fuzzing the parser across dialects, and confirmed each by lexing and parsing the string above through the public pipeline.

## Root cause

`trino.rs` builds Trino from `ansi::raw_dialect()`, then clears the keyword sets and repopulates them from `TRINO_UNRESERVED_KEYWORDS` / `TRINO_RESERVED_KEYWORDS`:

```rust
trino_dialect.sets_mut("unreserved_keywords").clear();
trino_dialect.update_keywords_set_from_multiline_string("unreserved_keywords", TRINO_UNRESERVED_KEYWORDS);
```

The inherited ANSI grammar still contains `Ref::keyword("TRANSIENT")`, `Ref::keyword("OVERWRITE")` and `Ref::keyword("SEQUENCE")` (from `TemporaryTransientGrammar`, `InsertStatementSegment`, and the alter/sequence grammar). `Dialect::expand` only creates a library entry for a keyword that is present in a keyword set, so these references resolve to nothing and `Dialect::ref` takes its `None` arm and panics. Trino does not override those statement segments, so the ANSI grammar with these refs is used as is.

## The tricky part, and why I did not just send a PR

The obvious quick fix, adding `TRANSIENT`, `OVERWRITE`, `SEQUENCE` to Trino's keyword sets, would be wrong. Trino does not support transient tables, `INSERT OVERWRITE`, or sequences, so registering those keywords would make the parser accept syntax that is not valid Trino.

I can see a few real fixes, and which one you want is a dialect design call, so I would rather ask than guess:

1. Override `CreateTableStatementSegment`, `InsertStatementSegment` and the alter/sequence grammar in Trino to drop the unsupported keyword references, matching what Trino actually accepts (and SQLFluff's Trino dialect).
2. Make `Dialect::ref` degrade gracefully when a grammar references a keyword that is not in the dialect's sets, so an optional keyword simply does not match rather than crashing. This would also guard any other dialect against the same class of panic, though it changes core behavior and might hide genuine dialect misconfigurations.

Happy to put up a PR for whichever direction you prefer.

dkxmercury:
Fixes #3017.

## Summary

Implements the four SQLFluff commits you listed, so the Trino dialect stops panicking on everyday statements and picks up the upstream grammar it had missed.

Before this, parsing any of these panicked instead of producing a tree or a clean error:

```
create table              -> Grammar refers to 'TRANSIENT' which was not found in the dialect
insert into t values (1)  -> Grammar refers to 'OVERWRITE' which was not found in the dialect
alter table t             -> Grammar refers to 'SEQUENCE' which was not found in the dialect
```

## What was ported

- **`c6b39aae`** `TemporaryTransientGrammar` becomes `Nothing`, which removes the inherited `TRANSIENT` reference. `PrimaryKeyGrammar`, `ForeignKeyGrammar`, `UniqueKeyGrammar` and `TemporaryGrammar` are silenced the same way, matching upstream.
- **`d36f2f06`** an explicit Trino `StatementSegment` listing only the statements Trino supports, plus Trino `CREATE TABLE`, `ColumnDefinitionSegment` and `TransactionStatementSegment`. Also adds the `FUNCTION` keyword, which that commit added to the Trino keyword list.
- **`74aaf399`** Trino `INSERT` without `OVERWRITE`, and `COMMIT`, `ROLLBACK` and `SET SESSION`.
- **`07db73d1`** Trino `ALTER TABLE`, covering RENAME, ADD/DROP/RENAME/ALTER COLUMN, SET AUTHORIZATION, SET PROPERTIES and EXECUTE, without unsupported inherited options.

Listing the supported statements explicitly is the part that actually closes off the panics, since unsupported ANSI branches such as sequences are no longer reachable and cannot resolve a keyword Trino does not define.

One addition beyond a direct port: `ALTER TABLE ... EXECUTE f(x => 1)` needs `=>` to lex as a single token, so the dialect now inserts a `fat_right_arrow` lexer matcher alongside the existing `right_arrow`, mirroring `StringLexer("fat_right_arrow", "=>", ...)` upstream. Without it the argument list is unparsable.

## Testing

- The three panics are gone. `create table`, `create table t (a integer)`, `insert into t values (1)`, `insert into t (a) select 1`, `alter table t rename to u`, `alter table t add column c integer`, `alter table t execute f(x => 1)`, `commit`, `rollback`, `start transaction isolation level serializable` and `set session foo = 1` all parse.
- Added the upstream fixtures for CREATE TABLE, ALTER TABLE, INSERT, COMMIT, ROLLBACK, START TRANSACTION and SET SESSION, with generated yml snapshots. All parse with no unparsable segments.
- Full `cargo test -p sqruff-lib-dialects --test dialects` run is clean, and I ran it twice before and after the change to be sure the result was stable. `cargo test -p sqruff-lib-core` is green, `cargo fmt --check` and `cargo clippy -p sqruff-lib-dialects --tests -- -Dwarnings` are clean.

One local note in case it helps anyone else on Windows: the fixture tests are sensitive to line endings, and with `core.autocrlf=true` a set of unrelated fixtures with multi line string bodies fails and the failing set varies between runs. With an LF checkout the suite is stable and green, which is what I used for the comparison above.


krutoileshii:
What's needed for this one?

PR #3097, opened 2026-08-12
fix(snowflake): parse multi-column unique constraints

### LINKED ISSUE #3048: Snowflake ALTER TABLE with multi-column unique constraint unparsable
### What Happened

The below Snowflake SQL is not recognized: Unparsable section. As a consequence, indentation rules are not applied properly (when set).
Note that the same ALTER TABLE with a single unique column works as expected.

### Expected Behaviour

Snowflake dialect recognize unique constraint with multiple columns.

### How to reproduce

```sql
ALTER TABLE my_table
ADD CONSTRAINT my_table_unique_idx UNIQUE (
  column_a, column_b
);

```

### Configuration
```ini
[sqlfluff]
dialect = snowflake
```

benfdking:
## Summary

- parse comma-delimited column lists in Snowflake `ALTER TABLE ... ADD CONSTRAINT ... UNIQUE (...)`
- add the exact regression fixture from #3048
- update the expected serialized parse tree

## Why

The Snowflake grammar accepted only one `ColumnReferenceSegment` inside the optional `UNIQUE` brackets. Valid composite unique constraints therefore produced an unparsable section and prevented indentation rules from applying correctly.

This aligns Sqruff with SQLFluff's out-of-line constraint grammar, which uses a delimited column-reference list.

Closes #3048.

## Validation

- `cargo fmt --check`
- `cargo test -p sqruff-lib-dialects --test dialects -- snowflake`
- `git diff --check`


benfdking:
The fixture isn't from sqlfluff so it should be put in the sqruff folder.

PR #3942, opened 2026-09-08
fix(ambiguous): enable AM02 for snowflake and trino (6369)

benfdking:
Summary:
- enable AM02 for Snowflake by removing it from the dialect skip set
- explicitly cover Snowflake and Trino in the AM02 dialect-support regression test
- regenerate rule documentation and advance the SQLFluff cursor exactly one first-parent SHA

Testing:
- hermetic Cargo workspace tests passed
- SQLFluff one-commit cursor verifier passed
- cargo fmt and git diff checks passed
- generated rule documentation check passed
- full Bazel suite passed: 49/49


benfdking:

PR #3977, opened 2026-09-09
fix(mysql): parse unscoped system variables (6439)

benfdking:
## Summary
- advance the SQLFluff cursor from fb1be11571a152b4cb1c9c589d9c7fdcc2f2cd39 to b9bcb9bc7134427a20a0d04b1e0edf55e7d5dafd
- allow unscoped MySQL system variables such as @@UNIQUE_CHECKS and accept system variables on SET assignment right-hand sides
- add the upstream SQL regression fixture, generated parse tree, and CV05 rule case

## Validation
- exact first-parent cursor guard
- focused MySQL dialect fixture suite
- CV05 rule fixture exercised through the Bazel workspace test
- cargo fmt --all -- --check
- git diff --check
- bazelisk test //... --test_output=errors (49/49 tests passed; three load-sensitive UI targets passed in isolation before the cached clean run)


benfdking:

PR #3988, opened 2026-09-09
feat(impala): support create table as select (6458)

benfdking:
Ports SQLFluff first-parent commit 80f4fc2d3bbc7839e41a9018ae4918118c309656 (Impala: support CREATE TABLE AS SELECT, #6458).\n\nSummary:\n- add a dedicated Impala CREATE TABLE AS SELECT statement grammar\n- support Impala CTAS table options before AS SELECT\n- add the upstream CTAS fixture and generated parse expectation\n- advance .sqlfluff-sha by exactly one first-parent commit\n\nValidation:\n- cargo fmt --all -- --check\n- git diff --check\n- direct Impala dialect fixture test passed\n- focused Bazel Impala dialect and cursor tests passed\n- bazelisk test //... --test_output=errors --jobs=4 completed all 49 targets: 47 passed; the playground and VS Code Playwright targets failed only under concurrent Rust build load\n- bazelisk test //playground:playwright_test --test_output=errors passed all 5 cases in isolation\n- bazelisk test //editors/code:playwright_test --test_output=errors passed all 6 cases in isolation\n\nChain base: port/sqlfluff-6453 (#3987).

PR #3989, opened 2026-09-09
fix(config): load extra config files last (6399)

benfdking:
Ports SQLFluff first-parent commit 4dc390f077ccef497c31e9421d73cd287891aa92 (Switch from appdirs to platformdirs, #6399).

Summary:
- load explicitly supplied config files after discovered configuration so they have highest precedence
- load the exact requested file, including arbitrary config filenames
- report a missing extra config path as a user-facing error
- add regression coverage for precedence, arbitrary filenames, and missing paths
- advance .sqlfluff-sha by exactly one first-parent commit

The Python appdirs-to-platformdirs dependency migration has no sqruff equivalent because sqruff does not depend on either package.

Validation:
- PYO3_PYTHON=/home/ben/Developer/sqruff/.venv/bin/python cargo test -p sqruff-lib core::config::tests -- --nocapture (22 passed)
- cargo fmt --all -- --check
- git diff --check
- bazelisk test //... --jobs=4 completed all 49 targets: 47 passed; the playground and VS Code Playwright targets failed only under concurrent Rust build load
- bazelisk test //playground:playwright_test --jobs=1 passed all 5 cases in isolation
- bazelisk test //editors/code:playwright_test --jobs=1 passed all 6 cases in isolation

Chain base: port/sqlfluff-6458 (#3988).


benfdking:

Issue #903, opened 2024-10-30
Do you support custom rules ?


Kiblyn11:
Do you support extending ruleset without rebuilding the binary ?
If not is it planned to do so ?

benfdking:
We haven't focused much on analyzing DDL statements, but I don't see why we wouldn't implement a generic rule like that. It seems like a fairly generic rule that people would use, "the ability to disallow certain types".

michaelvanstraten:
Hi @benfdking, really cool project. I am currently working on a migration squasher for a project, my flow currently looks like this:

1. Create a scratch db
2. run all migrations
3. run `pg_dump` on the scratch db using `std::process::Command`
4. hand stdout to `sqruff_lib_core::parser::lexer::Lexer::lex`
5. hand the tokens to `sqruff_lib_core::parser::Parser::parse`
6. massage the raw `ErasedSegments` (remove top-level sets by Postgres, remove `\restrict` and `\unrestrict`, remove comments, move `ALTER TABLE ONLY ... ADD CONSTRAINT ...` into the table creations)
7. dump it into a file.

I think if I just implement all these as `Linter` rules under a `squash` group, it would be much nicer to interact with. However, this requires a public API to set custom rules for the `Linter`. 

How do you feel about adding a `set_rulepack` to the linter struct? Would be a minimal change if you just document that calling `set_rulepack` twice or after `get_rulepack` will panic.

PR #4275, opened 2026-09-17
feat(linter): add a `set_ruleset` method to set rules on linter

michaelvanstraten:
This patch adds a `set_ruleset` to the `Linter` which basically allows to inject custom rule implementations into the linter.

I use this to add some project specific lints in addition to already existing lints. You could argue that this could have a nicer API and I would be willing to implement it if requested.

See also https://github.com/quarylabs/sqruff/issues/903#issuecomment-5699645055.


crates/lib/src/core/linter/core.rs
```diff
@@ -184,10 +184,16 @@ impl Linter {
...
     pub fn get_rulepack(&self) -> Result<RulePack, SQLFluffUserError> {
-        let rs = get_ruleset();
+        let rs = RuleSet::from_iter(rules());
         rs.get_rulepack(&self.config)
     }
 
+    pub fn set_ruleset(&self, rule_set: &RuleSet) -> Result<(), SQLFluffUserError> {
+        self.rulepack
+            .set(rule_set.get_rulepack(&self.config)?)
+            .map_err(|_| SQLFluffUserError::new("Can not call `Linter::set_ruleset` more then once or after `get_rulepack` has been called".to_string()))
```
michaelvanstraten, line 194: Linewrap?

benfdking (approved):

